PR #1, opened 2021-11-18
Added escrow example

cbisaillon:
Example of an escrow component, allowing two users to lock assets into a contract before accepting (or refusing) the trade.


basic/escrow/src/lib.rs
```diff
@@ -0,0 +1,136 @@
+use scrypto::prelude::*;
+
+blueprint! {
+    struct Escrow {
+        token_a: Vault,
+        token_b: Vault,
+        token_a_account: Account,
+        token_b_account: Account,
+        account_a_badge_ressource: ResourceDef,
+        account_b_badge_ressource: ResourceDef,
```
kroggen, line 10: ressource -> resource

cbisaillon: Thank you, my main language is french so sometimes I put two "s" instead of one :)

basic/escrow/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+            scrypto_assert!(self.can_add_tokens(user_id), "You have already added you tokens to the escrow");
+            scrypto_assert!(!self.trade_canceled, "The trade was canceled");
+
+            if user_id == self.account_a_badge.address() {
+                self.token_a.put(tokens);
+            } else if user_id == self.account_b_badge.address() {
+                self.token_b.put(tokens);
+            } else {
+                scrypto_abort("Not authorized !");
+            }
```
iamyulong, line 74: Given that `user_id` is matched against the `account_a_badge` and `account_b_badge` here, I believe the first 2 assertions and related methods are not needed

basic/escrow/tests/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+use radix_engine::transaction::*;
+use scrypto::prelude::*;
+
+#[derive(Copy, Clone)]
+struct User {
+    pub_key: Address,
+    address: Address
+}
+
+struct TestUtils<'a, L: Ledger> {
```
iamyulong, line 11: I like this `TestUtils`

iamyulong (commented):
Thanks for providing this example. It looks good to me and gives a lot of inspiration regarding RE transaction model.

I've left a few comments in case you want to improve it. I'm also happy to merge as-is.

basic/escrow/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        // Check if the user can add their tokens
+        fn can_add_tokens(&self, user_id: Address) -> bool {
+            (self.token_a.is_empty() && user_id == self.account_a_badge.address()) 
+            || (self.token_b.is_empty() && user_id == self.account_b_badge.address())
+        }
+
+        // Allow the users to put the tokens they want to trade inside the component's vault
+        pub fn put_tokens(&mut self, tokens: Bucket, badge: BucketRef) {
```
iamyulong, line 61: Have A second thought on this. We may also use the `auth` attribute.

```
#[auth(account_a_badge, account_b_badge)]
pub fn put_tokens(&mut self, tokens: Bucket) {
    let user_id = Self::get_user_id(auth);
}
```
Didn't test though.

basic/escrow/tests/lib.rs
cbisaillon: Thanks ! I would like to put it into a cargo package but I am not yet good enough with Rust to do it

basic/escrow/src/lib.rs
cbisaillon: Nice, I was trying to find a way to use the #[auth...] notation while being able to access the badge in the scope. That's what I was looking for !

cbisaillon:
Much cleaner now, thank @iamyulong

basic/escrow/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+                self.token_a.put(tokens);
+            } else {
+                self.token_b.put(tokens);
+            }
+        }
+
+        #[auth(account_a_badge)]
+        pub fn withdraw_all(&mut self) -> Bucket {
+            self.token_a.take_all()
+        }
```
iamyulong, line 60: I think there is a bug here. Consider the following even sequence:
1. User A deposits 500 DRG;
1. User B deposits 10 XRD;
1. User B accepts;
1. User A accepts;
1. User A withdraw all 500 DRG;
1. User A withdraw 10 XRD;
1. There is nothing left for B to withdraw.

Better to remove it.

cbisaillon: I forgot to remove "withdraw_all". It wasn't used anymore.

PR #3, opened 2021-11-22
Flashloan example

cbisaillon:


iamyulong:
Nice work!

It is a bit convolved but I think I've got your idea. 

The bit that confused me was the flow:
- An end user calls the `Caller` component
- The `Caller` calls the `FlashLoan` component
- The `FlashLoan` component calls the `Caller::execute(Bucket)->Bucket` method
- The `Caller` exploits a mocked opportunity 

I think we may be able to simplify it if the user calls directly the `FlashLoan` component, providing the address of the component which knows how to exploit opportunity.

The `FlashLoan` component requires the loan recipient to have a fixed interface (the `execute` method), which requires some adapter component like the `Caller`. Once Scrypto ABI is more stable and well-documented, we may make the flash loan accept any method.

cbisaillon:
@iamyulong I go inspired by how flashloans on Aave works: https://docs.aave.com/developers/guides/flash-loans. But you are right, we could remove the first step and directly call the `request_loan` method.

PR #4, opened 2021-11-23
Well looks like i did something here. Still all a learning process 

zDabski:


cbisaillon:
Thanks for submitting your examples but It seems to be the default code generated from `scrypto new-package`. I would suggest to add new things. Here are some examples:
- Change the data about the token (name, symbol, max_supply)
- Create a token with mutable supply instead of one with fixed supply
- Add authentication to this component. Maybe have two tokens. One that can be retrieved by anybody and the other that can only be retrieved by people who have the right badge

Happy learning !

zDabski:
Thanks for the response. Looks like I loaded the wrong one. This is the one
I meant to load. Does something like this work or there other editing that
must be done?
[image: image.png]

iamyulong:
Hi @zDabski, thank your for sharing your example and learning process!

I'm super happy to see you making process. However, I'm not going to merge this PR as it doesn't provide much info for others to learn.

I'd suggest you watch [Russ' workshop recording](https://www.youtube.com/c/RadixDLT/videos) and [Peirs and Matt's presentation](https://www.youtube.com/watch?v=Hn20w1gmpMw) to get some idea to code with.

[This website](https://defipulse.com/defi-list/) lists most DeFi applications, if you're interested.

PR #5, opened 2021-11-23
A basic example of an auction.

Aos-dev:


basic/auction/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+                highest_bid: Decimal::zero(),
+
+                auctioneer_badge: ResourceDef::from(Address::ResourceDef([0; 26])),
+                payment_claimed: false,
+            }
+            .instantiate();
+
+            // save the component address (it doesn't seem like there is a way to get it later from iside a method)
+            let mut auction_state = auction.get_state::<Self>();
+            auction_state.component_address = auction.address();
```
cbisaillon, line 63: Good idea ! I'll do something similar in my flashloan example to get the component's address

Aos-dev: This is more of a workaround. I thought it would be possible to get to the component address from the context or &self but unfortunately it doesn't seem so. At leat for the moment... maybe in the future version of Scrypto ;)

PostNutCIarity: Interesting. I tried making my own auction platform too. Wouldn't you want to create a vault for each bidder instead of each bid submission be pooled into one vault? That's my understanding of this, correct me if I'm wrong (I don't have a developer background).

Aos-dev: The tokens for bidding are expected to be fungible (otherwise it wouldn't be even possible to compare the bids) so there is no need to create a separate vault for every bidder. In the end it doesn't matter if you are returned exactly the same tokens that you deposited or not. Just as if you deposited $100 at one bank and then went to another one to withdraw it. I believe it doesn't matter to you and you don't even expect to get exactly the same $100 bill that you deposited. In the same time, an attempt by the bank to deliver you exactly the same $100 would trigger a massive, complicated process.
In the case of this auction component it wouldn't be that bad but still:

1. You would have to make sure that every bidder gets back the bid bond precisely from his vault
2. You would have to make sure that after an auction ends the auctioneer receives the bid bond precisely from the winner's vault (so you would also have to check who exactly won the auction, in the current logic it is enough if you just know the highest bid)
3. If you have 10'000 participants in the auction then you would spawn 10'000 vaults

And what would be the added value? Seems like none, so Occam's razor :)

PostNutCIarity: Thanks for the explanation this makes sense! :)

basic/auction/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+            scrypto_assert!(bid_bond.amount() == self.bid_bond, "Incorrect bid bond");
+
+            // take the bid bond
+            self.bid_bonds.put(bid_bond);
+
+            // mint a bidder badge
+            let bidder_badge = ResourceBuilder::new()
+                .metadata("name", "Bidder badge")
+                .metadata("auction", self.component_address.to_string())
+                .new_badge_fixed(1);
```
iamyulong, line 93: Once NFT is out, we will no longer need to create one resource per use.

basic/auction/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+
+            // save the bidder using the bidder badge resource definition
+            self.bidders.insert(bidder_badge.resource_def(), Bidder { bid: Decimal::zero(), bid_bond_reclaimed: false });
+
+            bidder_badge
+        }
+
+        pub fn bid(&mut self, bid: Decimal, bidder_badge: BucketRef) {
```
iamyulong, line 101: Given bidders are allowed to pose any bid without "collateral",  will need to be careful with issuing bidder badges. Maybe to only "trusted" users. :)

basic/auction/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+        bidders: LazyMap<ResourceDef, Bidder>,
+        highest_bid: Decimal,
+        
+        auctioneer_badge: ResourceDef,
+        payment_claimed: bool,
+    }
+
+    impl Auction {
+        pub fn new(offering: Bucket, duration: u64, payment_resource: Address, reserve_price: Decimal, bid_bond: Decimal) -> (Component, Bucket) {
+            scrypto_assert!(offering.amount() > Decimal::zero(), "Incorrect offering");
```
iamyulong, line 35: A minor heads-up: `scrypto_assert!` will be replaced with `assert!`.

iamyulong (approved):

PR #8, opened 2021-11-24
adding time based example

croquelois:
the creator of the instance receive a badge which allow him to withdraw token from the vault.

The vault is accumulating token at a rate specified during the creation of the instance.


iamyulong (approved):
LGTM

PR #9, opened 2021-11-25
Add a new example TokenSale blueprint

backyard-coder:
This commit adds a component that implements a simple token sale mechanism. A component like this could be usable on launch
platforms that sell new project tokens to a select group of whitelisted users.


iamyulong (approved):
Thanks for providing this example with detailed documentation!

PR #10, opened 2021-11-26
Add shared-company example 

MiroLiebschner:
The example of a company governed by its shareholders. More information in the Readme.

Hope this example makes into the repo :D If there is a specific reason why it won't make it, please write a comment so I can fix the issue.

Cheers, Miro

### DIFF
--- a/other/shared_company/src/lib.rs
+++ b/other/shared_company/src/lib.rs
@@ -1,0 +1,110 @@
+use scrypto::prelude::*;
+mod proposal;
+use crate::proposal::Proposal;
+
+/*
+A SharedCompany which is a company owned and governed by its shareholders.
+Shares can be bought (fixed rate for now).
+Funds can be spend using [Proposal] if accepted by a majority (of company_voting_tokens which are distributed with shares).
+This is only a demonstration. There are known security flaws in this design.
+(like: Buy a lot of shares, make proposal to send you all company_radix, vote for it).
+*/
+blueprint! {
+
+struct SharedCompany {
+    company_radix:  Vault,
+    company_shares: Vault,
+    company_voting_token: Vault,
+    share_counter: Decimal,
+    price_share: Decimal,
+    proposal_list: Vec<Address>,
+}
+
+
+impl SharedCompany {
+    // Creates a new component
+    pub fn new(price_share: Decimal) -> Component{
+
+
+        // shares are equity in the company. They can be bought for a fixed rate or returned for a part of the company vault
+        let shared_company_share_resource_def = ResourceBuilder::new()
+        .metadata("name", "SharedCompany share").metadata("symbol", "SC")
+        .new_token_fixed(1_000_000);
+
+        // voting tokens are used to vote on proposals
+        let shared_company_voting_token_resource_def = ResourceBuilder::new()
+        .metadata("name", "SharedCompany voting token").metadata("symbol", "SCVT")
+        .new_token_fixed(1_000_000);
+
+        // stores the addresses of proposal-components
+        let proposal_address_list: Vec<Address>= Vec::new();
+
+
+    //populate the SharedCompany struct and instantiate a new component
+    Self {
+        company_shares: Vault::with_bucket(shared_company_share_resource_def),
+        company_voting_token: Vault::with_bucket(shared_company_voting_token_resource_def),
+        company_radix: Vault::new(RADIX_TOKEN),
+        share_counter: Decimal(0.0 as i128),
+        price_share: price_share,
+        proposal_list: proposal_address_list
+    }.instantiate()
+
+    }
+
+    /// buys an amount of shares and returns change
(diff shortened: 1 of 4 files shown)


other/shared_company/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        // stores the addresses of proposal-components
+        let proposal_address_list: Vec<Address>= Vec::new();
+
+
+    //populate the SharedCompany struct and instantiate a new component
+    Self {
+        company_shares: Vault::with_bucket(shared_company_share_resource_def),
+        company_voting_token: Vault::with_bucket(shared_company_voting_token_resource_def),
+        company_radix: Vault::new(RADIX_TOKEN),
+        share_counter: Decimal(0.0 as i128),
```
cbisaillon, line 48: Great example ! You can use `Decimal::zero()`

other/shared_company/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    }
+
+    /// buys an amount of shares and returns change
+    pub fn buy_shares(&mut self, payment: Bucket) -> (Bucket, Bucket, Bucket) {
+        let max_share_buy_power = payment.amount() / self.price_share;
+        // Increase the share_counter so the amount of shares that are outstanding is tracked
+        self.share_counter += max_share_buy_power;
+        // take our price in XRD out of the payment and store it in the company vault
+        let collected_xrd = payment.take(self.price_share * max_share_buy_power);
```
cbisaillon, line 61: Isn't this like putting the full payment bucket ?

cbisaillon: I think you would need to round down the result on line 57

other/shared_company/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        self.company_radix.put(collected_xrd);
+        // return the share(s) and change
+        (self.company_shares.take(max_share_buy_power), payment, self.company_voting_token.take(max_share_buy_power))
+    }
+
+     /// sells an amount of shares and for a part of the companies xrd
+     pub fn sell_shares(&mut self, shares: Bucket, voting_token: Bucket) -> (Bucket, Bucket) {
+        if shares.amount() != voting_token.amount()  {return (shares, voting_token)};
```
cbisaillon, line 69: I think using the same token for shares and voting power would reduce the complexity

cbisaillon: You could also do something like this:
`scrypto_assert!(shares.amount() == voting_token.amout(), "shares_amount must equal voting_token amount")`

other/shared_company/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        // calculates the percentage of all shares
+        let percentage_of_all_shares = shares.amount() / self.share_counter;
```
cbisaillon, line 71: You don't have to keep track of the share count. You can store the resourceDef of the shares in the component's data then do `self.share_resource_def.supply()`

cbisaillon, line 71: You don't have to keep track of the share count. You could just save the share's resource_def in the component's data. Then do something like `self.share_resource_def.supply() - self.company_share.amount()`

other/shared_company/src/proposal.rs
```diff
@@ -0,0 +1,117 @@
...
+
+                     // The token that the user gets in exchange for their "no" voting.
+                     let replacement_token_no_resource_def = ResourceBuilder::new()
+                     .metadata("name", "Replacement token no").metadata("symbol", "RTN")
+                  .new_token_fixed(1_000_000);
+
+                    // the vault that holds the costs that are associated with the proposal
+                    let cost_vault = Vault::new(cost.resource_def());
+                    // fills the cost vault
+                    cost_vault.put(cost);
```
cbisaillon, line 45: You can do it in one line:

`let cost_vault = Vault::with_bucket(bucket)`

other/shared_company/src/proposal.rs
```diff
@@ -0,0 +1,117 @@
...
+
+                  Self {
+                    cost_vault: cost_vault,
+                    destination_adress_funds: destination_adress_funds, reason: reason,
+                    company_voting_token : Vault::new(company_voting_token_resource_def),
+                    replacement_tokens_type_yes: Vault::with_bucket(replacement_token_yes_resource_def),
+                    replacement_tokens_type_no: Vault::with_bucket(replacement_token_no_resource_def),
+                    yes_counter: Decimal(0.0 as i128),
```
cbisaillon, line 53: Instead of creating two tokens (Yes and No), why not create a single Vote badge. When the user shows their vote badge they say wether they vote yes or no. Then the component burns the badge so that they can only vote once

iamyulong (approved):
Thanks for providing this example!

PR #12, opened 2021-11-27
Book Library Example

charliecooper45:
This example implements a Book Library where users can pay for membership and then borrow books. 

It's a bit abstract as I wanted to build something easy to reason with as I'm completely new to both Rust and DeFi!

Any feedback is greatly appreciated. 

Thanks

### DIFF
--- a/basic/digital-library/src/lib.rs
+++ b/basic/digital-library/src/lib.rs
@@ -1,0 +1,151 @@
+use sbor::*;
+use scrypto::prelude::*;
+
+#[derive(TypeId, Encode, Decode)]
+struct Book {
+    title: String,
+    author: String,
+}
+
+blueprint! {
+    struct Library {
+        // librarian (admin) badge
+        librarian_badge: ResourceDef,
+
+        // holds the library funds, from late fees and memberships
+        fees: Vault,
+
+        // maps ISBN numbers to books
+        books: HashMap<String, Book>,
+
+        // each member receives a badge
+        member_badges: Vault,
+        member_badge_def: ResourceDef,
+
+        // the cost of membership to the library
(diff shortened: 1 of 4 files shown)


basic/digital-library/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+            let book_two = Book { author: String::from("Frank Herbert"), title: String::from("Dune") };
+            let book_three = Book { author: String::from("Dan Abnett"), title: String::from("Horus Rising") };
+            books.insert(String::from("9781611297560"), book_one);
+            books.insert(String::from("9780450011849"), book_two);
+            books.insert(String::from("9781844162949"), book_three);
+
+            let member_badges_bucket = ResourceBuilder::new()
+                .metadata("name", "Library Membership Badge")
+                .metadata("symbol", "LMB")
+                .new_badge_fixed(member_badge_count);
```
cbisaillon, line 63: I think it would be cool for this badge to be mutable. You would create another single badge "member_badge_minter", stored inside a vault of the component. Then inside `register`, you mint a new badge only when a user needs to register. I think this will make it more optimized to get the total number of members. Instead of having to count each keys in the hashmap, you could do `member_badges_resource_def.supply()`

cbisaillon:
Really great example ! I added a comment about having the member badges mutable. I think it would be a great addition. 

Also, it might be overkill but I think it would be cool if you added a field "value" on the book struct. When users want to borrow a book, they would have to deposit the value of the book + the fees. When they return the book, they get the value of the book + the fee (if it's not overdue). This way, it would incentivize users not to steal the books and you would only have one `return_book` method

PR #13, opened 2021-11-28
Add transit example

shanesreal:
This is a basic example of a transit, accounts can purchase ticket tokens (minted) and go on rides (burned). Accounts with a host badge can withdraw/collect payments and also shutdown the rides. Transit rides are tracked using riders public key and the current epoch.

### DIFF
--- a/basic/transit/src/lib.rs
+++ b/basic/transit/src/lib.rs
@@ -1,0 +1,144 @@
+use scrypto::prelude::*;
+
+blueprint! {
+    struct Transit {
+        /// Transit rider public key/epoch
+        riders: LazyMap<Vec<Address>, u64>,
+        /// Ticket Resource Definition and Mint/Burn Authorization
+        ticket_resource_def: ResourceDef,
+        ticket_minter: Vault,
+        /// Enable/Disable rides
+        american_rides: bool,
+        european_rides: bool,
+        /// Ticket/Ride prices
+        ticket_price: Decimal,
+        ride_price: Decimal,
+        /// Control rides/Withdraw payments
+        american_host_badge: ResourceDef,
+        european_host_badge: ResourceDef,
+        /// Collected revenue from ticket sales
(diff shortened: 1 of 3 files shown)


basic/transit/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+        ) -> (Component, Bucket, Bucket) {
+
+            let valid_cli_arguments = price_per_ticket > 0.into() && price_per_ride > 0.into();
+            scrypto_assert!(valid_cli_arguments, "Invalid CLI arguments");
+
+            let american_badge = ResourceBuilder::new()
+                .metadata("name", "American Host Badge")
+                .metadata("symbol", "APB")
+                .metadata("description", "A badge that grants american host privileges")
+                .new_token_fixed(1);
```
cbisaillon, line 39: Why not use new_badge_fixed instead of creating tokens ?

cbisaillon (commented):
Good job, great example !

basic/transit/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+
+            // Keep track of riders public key/epoch
+            if self.riders.get(&Context::transaction_signers()) == None {
+                info!("Hi, this is your first ride on a transit, have fun!");
+                self.riders.insert(Context::transaction_signers(),Context::current_epoch());
+            } else if self.riders.get(&Context::transaction_signers()).unwrap() == Context::current_epoch() {
+                info!("Hi, you have already used the transit more than once during epoch: {}", Context::current_epoch())
+            } else {
+                info!("Hi, welcome back, you have not used the transit during epoch: {}", Context::current_epoch());
+                self.riders.insert(Context::transaction_signers(),Context::current_epoch());
```
iamyulong, line 140: This would be perfect candidate for non-fungible NFT. You issue an NFT for each rider and record the epoch time inside the NFTs.

PR #14, opened 2021-11-28
added example with a crowdsourcing campaign

Kansuler:
This is a basic example where you can start a crowdsourcing campaign to raise funds.

The rules that binds this contract is that the fundraiser instantiated the component with a funding goal that needs to be reached within a given duration of epochs.

Any wallet can pledge towards the campaign to become a patron and receive a mutable patron badge. The patron badge is unique and can be used to recall a pledge as long as the campaign is either ongoing, or the campaign goal hasn't been met.

If the campaign is finished and the goal has been achieved, the fundraiser can withdraw all collected funds.


defi/crowdsourcing/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+            refund
+        }
+
+        /*
+        As fundraiser, withdraw collected XRD if goal has passed, and the last_epoch has passed.
+        */
+        #[auth(fundraiser_badge_def)]
+        pub fn withdraw(&self) -> Bucket {
+            assert!(Context::current_epoch() > self.last_epoch, "campaign has not ended yet.");
+            assert!(self.collected_xrd.amount() > self.goal, "campaign did not reach it's goal.");
```
cbisaillon, line 116: I think you want to use ">=" instead of ">", or else you won't be able to withdraw on the end epoch

cbisaillon (commented):
Good job !

defi/crowdsourcing/src/lib.rs
Kansuler: Oh, good point! Thanks, I'm fixing in a moment.

iamyulong:
Really good example. I'm going to merge it now, but feel free to make another PR for enhancements.

PR #17, opened 2021-12-02
Add an airdrop component

Fanti77:
This PR adds an example for an airdrop component that may be used to distribute tokens to users.

### DIFF
--- a/basic/airdrop/src/lib.rs
+++ b/basic/airdrop/src/lib.rs
@@ -2,21 +2,29 @@ use scrypto::prelude::*;
 
 blueprint! {
     struct Airdrop {
+        admin_badge: ResourceDef,
         recipients: Vec<Address>,
     }
 
     impl Airdrop {
-        pub fn new() -> Component {
-            Self {
-                recipients: Vec::new(),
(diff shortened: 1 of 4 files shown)


basic/airdrop/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    impl Airdrop {
+        pub fn new() -> Component {
+            Self {
+                recipients: Vec::new(),
+            }
+            .instantiate()
+        }
+
+        pub fn add_recipient(&mut self, recipient: Address) {
```
cbisaillon, line 16: Really nice example ! I think badges for authentication would be a good use case for this component. Allowing only people presenting a certain badge to call `add_recipient` and `perform_airdrop`

Fanti77: Thanks for the feedback! Initially I wanted to keep the example as lean as possible. When you think about it, though, it makes sense for all examples to be properly secured. I have now added authorization to both methods as per your suggestion.

iamyulong:
Thanks for contributing to this community Scrypto repo. It looks very good to me.

PR #18, opened 2021-12-03
feat: Added Virtual XRD example

bdebenon:


iamyulong:
Nice one!  Virtual XRD 100% backed by XRD. I like it.

PR #20, opened 2021-12-07
Insurance

BlindDev:
Simple Insurance example. 

An org is able to create different policies with specific price, coverage and supply. The required coverage will be locked
An insurer can then purchase policy by address

The org can approve a payment of valid policies to the insurer. 
Unused policies or expired purchases can be burned

### DIFF
--- a/defi/insurance/src/lib.rs
+++ b/defi/insurance/src/lib.rs
@@ -1,0 +1,250 @@
+use scrypto::prelude::*;
+
+
+blueprint! {
+    struct Insurance {
+        /// Mint authorization to TL tokens.
+        org_vault: Vault,
+        org_badge: ResourceDef,
+
+        // Non locked assets
+        assets_pool: Vault,
+        // Locked assets by policies
+        locked_pool: Vault,
+       
+        // HashMap of policy address and details
+        policies: HashMap<Address, Vault>,
+        // HashMap of insurer and policy vaults HashMap
+        purchases: HashMap<Address, HashMap<Address, Vault>>,
+    }
+
+    impl Insurance {
+        
+        // Create new Insurance component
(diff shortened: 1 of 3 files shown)


defi/insurance/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+            // take payment
+            let payment = bucket.take(price);
+            self.assets_pool.put(payment);
+
+            // return the rest bucket
+            bucket
+        }
+
+        // Approve payment with specific amount to the insurer using policy address 
+        //#[auth(org_badge)]
```
cbisaillon, line 143: Did you forget to remove the comments on the `auth` macros ?

BlindDev: @cbisaillon 
Ah, yes. I have to leave those commented because otherwise  my VSCode highlights everything red with following error (make_policy example):

```
expected 6 arguments, found 7rust-analyzermismatched-arg-count
scrypto_derive
pub macro auth
```

```
expected 6 arguments, found 7rust-analyzermismatched-arg-count
out::Insurance
pub fn make_policy(&self, arg0: String, arg1: Decimal, arg2: Decimal, arg3: u64, arg4: Decimal, auth: ::scrypto::resource::BucketRef)
out::blueprint::Insurance
pub fn make_policy(&mut self, policy_type: String, coverage: Decimal, price: Decimal, duration: u64, supply: Decimal)
```

I wasn't not sure is this something related to my rust-analyzer setup or some exceed settings, so I ended up with commenting auth macros everywhere.

iamyulong:
Thanks for submitting this example. It looks good to me. Also curious to know if auth would work after removing the comment.

BlindDev:
The auth badge works in general, it's just always highlighted as an error in my VSCode. I will improve my code in the next PR so it will be uncommented. Thanks!

PR #21, opened 2021-12-09
Donations example

BlindDev:
In this blueprint  example an admin can create Donations component with some fee in %
A user can create any amount of different badges for donators. 
Donators can then purchase badges at corresponding price.

### DIFF
--- a/defi/donations/Cargo.toml
+++ b/defi/donations/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "donations"
+version = "0.1.0"
+edition = "2018"
+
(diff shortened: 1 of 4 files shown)


iamyulong:
It seems that anyone is allowed to create a badge for any "owner". Is this by design? 

The double admin badge (one owned by component, one by external actor) is very cool.

BlindDev:
> It seems that anyone is allowed to create a badge for any "owner". Is this by design?

Yep, it was just a first version, I have a few more updates for this, such as owner registration request and approvement by an admin

PR #22, opened 2021-12-10
Adds a TokenBurn component

backyard-coder:
This example illustrates how a component may be used to effectively burn tokens. This works as the Scrypto implementation currently allows dangling vaults, to which no entity holds a reference. See https://github.com/radixdlt/radixdlt-scrypto/issues/87

### DIFF
--- a/basic/token-burn/src/lib.rs
+++ b/basic/token-burn/src/lib.rs
@@ -1,0 +1,28 @@
+use scrypto::prelude::*;
+
+blueprint! {
+
+    struct TokenBurn {
+        token_address: Address,
+        burnt_amount: Decimal,
+    }
+
+    impl TokenBurn {
+        pub fn new(token_address: Address) -> Component {
+            Self {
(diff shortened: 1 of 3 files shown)


basic/token-burn/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+            Self {
+                token_address: token_address,
+                burnt_amount: Decimal::zero(),
+            }
+            .instantiate()
+        }
+
+        pub fn burn_tokens(&mut self, tokens_to_burn: Bucket) {
+            self.burnt_amount += tokens_to_burn.amount();
+            Vault::new(ResourceDef::from(self.token_address)).put(tokens_to_burn);
```
iamyulong, line 21: I like the idea of allowing people to "effectively" burn any tokens. Although, the current implementation may not work once dangling vault check is enforced. After that, will always need to attach the vaults to a component.

PR #24, opened 2021-12-15
scrypto-bonding bonding curve amm initial release

devmannic:
An AMM based on bonding curves.  Enjoy!

See the `README.md` for details.  Note this is compatible with `main` branch pre-Alexandria (right now!) :).  Some minor changes will be needed here and to one of the dependency libraries (which I'm also the author of)

### DIFF
--- a/defi/amms/bonding/README.md
+++ b/defi/amms/bonding/README.md
@@ -1,0 +1,50 @@
+# scrypto-bonding
+
+srypto-bonding: A Scrypto package for creating and using bonding curves as an automated market maker on Radix
+
+The main blueprint to interact with is `BondingAMM`.  It has buy/sell functionality against a single
+bonding curve defined by a plugable second component.  It will instantiate a component from the `RatioBondingCurve`
+blueprint if no other curve component is provided.
+
+There is also an extrememly simple `BasicBondingCurve` blueprint (more like a flat line) which could be used to implement
+a simple "wrapped" or "virtual" token from another.  Or it's a a good template for making your own curve component.
+
+Some useful utilities are included too.  An internal crate is used to provide the `blueprint_stub` macro which
+automates creating stub functions from a trait so calling another component is ergonomic.  Also included and
+used for the `RatioBondingCurve` is a reusable arbitrary precision number implmentation that converts to/from Decimal
+It is precise but not yet optimized.  Bounded (BigInt) or unbounded (BigRational) precision is configurable with a feature flag.
+
+## Bonuses:
+
+* This package is well tested (though not completely) with both unit and integration tests.  Integration
+testing is not as easy as it should be, but using the [`scrypto-unit`] library helps a whole lot
+
+* Key portions of `BondingAMM` are implemented using [`scrypto_statictypes`] which provides compile-time resource
+checking and additional runtime checking for enhanced producitivity and safety.
+
+[`scrypto-unit`]: https://github.com/plymth/scrypto-unit
+[`scrypto_statictypes`]: https://github.com/devmannic/scrypto_statictypes
+
+## References:
+
+If you want to read more about bonding curves here some things that might be useful:
+
+Reading List
+* https://blog.relevant.community/bonding-curves-in-depth-intuition-parametrization-d3905a681e0a?gi=8df5128ab916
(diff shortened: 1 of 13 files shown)


backyard-coder:
Very nice example! 👍 
Love the comprehensive documentation!

I have one question, though: what is the rationale behind integrating the bonding curve as a component/service rather than as a trait with different implementations? I see that it illustrates nicely how inter-blueprint calls can be made. But is there a reason beyond that? If I did not miss anything substantial, the bonding curve simply makes some stateless calculations (i.e. only based on input parameters). Would it not have been easier to implement the calculations inside the same blueprint?

devmannic:
> I have one question, though: what is the rationale behind integrating the bonding curve as a component/service rather than as a trait with different implementations?

@backyard-coder  That's a quite reasonable question to ask.  It could have absolutely been merged all together.  I'm sure there's more reasons I could come up with but the summary of reasons why I went this route is:

1. I wanted the token lifetime/transfer logic and AMM API to be reusable and easy to understand since that's most of what distinguishes an implementation in Scrypto from one in say, Solidity
2. The actual bonding curve math can take many forms, and they can now be implemented by someone else not attached to the AMM's blueprint immutability (when deployed).
    a. One could implement support for the same style curves I did, but use way more optimized mathematics (as is done in the Bancor implementation in Solidity)
    b. Or one could use different curve formulas all together (for example a sigmoid curve instead of exponential)
3. It's now not too hard to extend this "separable" implementation to use two diffferent curves for buy vs sell or to allow an authorized user to update the curve in use if desired.
4. If this becomes the basis for production code, having separable components makes it easier for security audits and independent upgrades.

defi/amms/bonding/README.md
```diff
@@ -0,0 +1,50 @@
...
+* https://discourse.sourcecred.io/t/bonding-curve-references/271
+
+Code
+* https://github.com/oed/bonding-curves/blob/master/contracts/PolynomialCurvedToken.sol
+* https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7b7ff729b82ea73ea168e495d9c94cb901ae95ce/contracts/math/Power.sol
+* https://github.com/bancorprotocol/contracts-solidity/blob/c9adc95e82fdfb3a0ada102514beb8ae00147f5d/solidity/contracts/converter/BancorFormula.sol
+
+## Notes:
+
+The initial release is pre-Alexandria with expected (minimal) API incompatabilities
```
iamyulong, line 47: Would you be interested to update this example to work with Alexandria release?

defi/amms/bonding/bonding_macros/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+        use super::*;
+        blueprint! {
+            struct #ident {}
+            #the_impl
+        }
+        }
+        // reexport the stub only
+        pub use #mod_name::#ident;
+    })
+}
```
iamyulong, line 107: This is an interesting idea - transforming a trait into a blueprint with empty implementation.  One suggestion if I can make is probably to use `#(#var)*` syntax from `quote` so you can avoid generating-and-parsing token stream.

iamyulong (commented):
Really good example of complex Maths.

iamyulong:
Merging it now, but feel free to make further PRs for enhancements.

defi/amms/bonding/bonding_macros/src/lib.rs
devmannic: I really liked this concept of transforming the trait, but absolutely the implementation is just a first cut to have it work.  I took the approach of doing the least intrusive thing to reuse the most code, but I'm not sure the impact on the size of the generated code, if the release build really would be smart enough to remove all of the non-pub stuff during LTO, particularly because of the `extern "C"` it might not.  Also I didn't want to depend too much on the implementation of blueprint knowing it could change.

But if this kind of thing were part of Scrypto I would imagine that maybe a better way would be to duplicate some code from the existing `blueprint!` macro but only generate the minimal amount of API stubs, and not actually make the entire blueprint.  It could also generate a real Rust trait in such a way that a blueprint could implement it to get static checks that the "receiving" side of the function/method calls have all the right methods.

I'll open an enhancement issue to track this idea.

PR #26, opened 2021-12-16
Fisherman Game example

BlindDev:
Each player tries to capture bigger fish (in grams for this particular example), the bigger the depth value (1-10) the bigger fish can be captured, but the chances of winning will be reduced correspondingly. Picked depth will multiply the hash value so the risk may bring the winning.
Admin finishes the game and results will be calculated. The winner takes the prize pool minus the fee. If there are no winners, the pool will be used in the next game.

### DIFF
--- a/other/fisherman/Cargo.toml
+++ b/other/fisherman/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "fisherman"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+sbor = { path = "/Users/woof/Developing/Projects/Blockchain/radix/radixdlt-scrypto/sbor" }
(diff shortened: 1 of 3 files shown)


other/fisherman/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "fisherman"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+sbor = { path = "/Users/woof/Developing/Projects/Blockchain/radix/radixdlt-scrypto/sbor" }
```
cbisaillon, line 7: Can you make this example compatible with scrypto v0.2.0 please ?

BlindDev: @cbisaillon 

Done

Thanks!

cbisaillon:
Thank you !

PR #35, opened 2021-12-22
Updated basic/transit to use Scrypto v0.2.0

cbisaillon:
@shanesreal Can you take a look and tell me if it's ok ?


shanesreal:
> @shanesreal Can you take a look and tell me if it's ok ?

@cbisaillon I just tested these changes using the latest simulator v0.2.0 on my local machine and everything seems to be working ok, thank you!

PR #38, opened 2021-12-22
Updated toys/accumulator to use Scrypto v0.2.0

cbisaillon:
@croquelois Can you take a look and tell me if this is ok ?


croquelois:
tested locally, everything work fine, good to merge !

PR #41, opened 2021-12-22
Updated defi/crowdsourcing to use Scrypto v0.2.0

cbisaillon:
@Kansuler Can you take a look and tell me if this is ok ?


Kansuler:
Thanks for the updates, it looks great. 👍

PR #48, opened 2021-12-22
Updated defi/virtual_xrd to use Scrypto v0.2.0

cbisaillon:
@bdebenon Can you take a look and tell me if this is ok ?


bdebenon:
Just tested this locally and it runs as expected. Changes also look good. Thank you for updating this! 👍 

We should be good to merge this!

PR #45, opened 2021-12-22
Updated defi/marketplace to use Scrypto v0.2.0

cbisaillon:
@machisuji Can you take a look and tell me if this is ok ?

PR #49, opened 2021-12-23
Marketplace NFT update

machisuji:
Expanding on @cbisaillon's proposal in #45 this updates the marketplace example such that it uses NFTs for the order tickets.

While at it I also refactored the code a bit to remove the unnecessary separation of buy and sell orders and the resulting code duplication.


cbisaillon:
I think the problem with the demo.sh script is that the format return from `resim new-account` `resim show [account]` has changed since v0.2.0

machisuji:
@cbisaillon All done now! NFTs are used for the order tickets and the demo has been updated accordingly too. Also the demo bash and powershell scripts got out of sync. So now I've refactored the demo so that both shell and powershell actually use the same file for the demo steps themsevles. This way they only have to changed in one place (steps.ps1).

cbisaillon:
Good job. This is a high quality example !

cbisaillon:
@machisuji I suggest submitting this example to the Scrypto ecosystem page: https://developers.radixdlt.com/ecosystem

machisuji:
Thanks @cbisaillon ! Will do.

PR #52, opened 2021-12-28
Airdrop with fixed token by recipient

yetininoliviercoulibaly:
Hello all,

Thank you for giving us the opportunity to participate in this enriching adventure that is radix 
I take advantage of my learning of scrypto and defi for create this component which is largely inspired by the airdrop component. The fundamental difference is that the number of tokens of each recipient is passed as a parameter when created the component. The caller must provide a sufficient number of tokens for all recipients.
I'm waiting for your remarks and suggestions.  

Best regards,


basic/airdrop-with-fixed-tokens-by-recipient/src/lib.rs
```diff
@@ -0,0 +1,68 @@
+use scrypto::prelude::*;
+
+
+blueprint! {
+
+    struct AirdropWithFixedTokensByRecipient {
```
cbisaillon, line 6: This example looks too much like this one: https://github.com/radixdlt/community-scrypto/tree/main/basic/airdrop.
Would you be able to make it more different ? For example, instead of directly sending tokens to the users, allow them to withdraw the tokens by specifying their badges.

basic/airdrop-with-fixed-tokens-by-recipient/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        pub fn get_minimal_required_tokens(&self) -> Decimal {
+            let recipients_count = self.recipients.len();
+            return Decimal::from(recipients_count) * self.tokens_by_recipient; 
+        }
+
+        #[auth(admin_badge)]
+        pub fn add_recipient(&mut self, recipient: Address) {
+                self.recipients.push(recipient);
```
cbisaillon, line 34: You should add a check to make sure the recipient is not already inside the `self.recipients` vec. You can use `assert!`

basic/airdrop-with-fixed-tokens-by-recipient/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+            
+            let recipients_count = self.recipients.len();
+            assert!(recipients_count > 0, "You must register at least one recipient before performing an airdrop");
+
+            let required_tokens = self.get_minimal_required_tokens();
+            assert!(tokens.amount() >=required_tokens, "The tokens quantity is not sufficient"); 
+
+            info!("perform_airdrop with recipients_count {} required_tokens {}", recipients_count,  required_tokens);
+
+            for i in 0..recipients_count {
```
cbisaillon, line 48: Here you actually don't have to loop twice over the recipients. You can just send `tokens.amount() / recipient_count`

basic/airdrop-with-fixed-tokens-by-recipient/src/lib.rs
yetininoliviercoulibaly: Hello Clement, 
Thanks for your review. 
Indeed, the 2 implementations are very close.
I will carry out the implementation you suggest.

basic/airdrop-with-fixed-tokens-by-recipient/src/lib.rs
yetininoliviercoulibaly: Ok, I will make it

basic/airdrop-with-fixed-tokens-by-recipient/src/lib.rs
yetininoliviercoulibaly: Indeed, I will change it.

basic/airdrop-with-fixed-tokens-by-recipient/src/lib.rs
yetininoliviercoulibaly: @cbisaillon I just updated the source code. could you check please ?

basic/airdrop-with-withdraw/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+            return required_tokens;
+        }
+
+        #[auth(admin_badge)]
+        pub fn get_required_tokens(&self) -> Decimal {
+            return self.get_required_tokens_internal();
+        }
+
+        #[auth(admin_badge)]
+        pub fn add_recipient(&mut self, recipient: Address, tokens : Decimal)  {
```
cbisaillon, line 45: You could accept a `Bucket` here instead of Decimal. That way, you can update the `recipients` map and `tokens` vault in the same method

basic/airdrop-with-withdraw/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+        fn get_required_tokens_internal(&self) -> Decimal {
+            let mut required_tokens  = Decimal::zero();
+            for val in self.recipients.values() {
+                required_tokens = required_tokens + *val ;  
+            }
+            return required_tokens;
+        }
+
+        #[auth(admin_badge)]
+        pub fn get_required_tokens(&self) -> Decimal {
```
cbisaillon, line 40: I don't understand the use of this method ? Why not just make sure the amount of token >= fixed_amount when calling `add_recipient` ?

basic/airdrop-with-withdraw/src/lib.rs
cbisaillon, line 45: Using the recipient address is not the resource oriented way. I suggest using badges. In the `new` function, create a new token for the badge:
```
// Used to mint new user badges
let minter_badge = ResourceBuilder::new_fungible(DIVISIBILITY_MINIMUM)
                                   .initial_supply_fungible(1); // Store this in a vault on the component

let user_badge_def = ResourceBuilder::new_fungible(DIVISIBILITY_MINIMUM)
                               .flags(MINTABLE)
                               .badge(minter_badge.resource_address(), MAY_MINT)
                               .no_initial_supply(); // Store this resource def on the component
```

Then, inside `add_recipient`, you would mint a new `user_badge` like this: 
```
let user_badge = self.minter_badge_vault.authorize(|auth|{
  self.user_badge_def.mint(1, auth)
});
```

basic/airdrop-with-withdraw/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        pub fn  get_recipient_available_tokens(&mut self, recipient_address : Address) -> Decimal
+        {
+            assert!(self.recipients.contains_key(&recipient_address), "Recipient not found");
+            let recipient_tokens  = self.recipients[&recipient_address]; 
+            info!("available_tokens : {}",recipient_tokens);
+            return recipient_tokens;
+        }
+
+        pub fn withdraw_token(&mut self, recipient_address : Address) -> Bucket
```
cbisaillon, line 72: After creating badges, here you can remove the `recipient_address` param and add a `#[auth(user_badge)]` macro on top of the method. You would then access the address of the badge like this in the method: `auth.resource_address()`

basic/airdrop-with-withdraw/src/lib.rs
cbisaillon: You would add a parameter to the `new` function to allow the person instantiating the component to specify a minimum amount

cbisaillon (changes requested):
Hi, here are some suggestions

basic/airdrop-with-withdraw/src/lib.rs
yetininoliviercoulibaly: Hi, thanks for the review.  This allowed me to better understand the philosophy of scrypto. 
But However I have a problem in the unit tests (try_withdraw_after_added_recipients_must_be_succeeded ) when calling the withdraw_token method . I have systematically **Insufficient balance** error. Although a deposit of user_badge was made when calling the method add_recipient on the account.  Could you tell me where my error could come from?

basic/airdrop-with-withdraw/tests/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+
+
+#[test]
+fn try_put_airdrop_tokens_with_sufficient_tokens_must_be_succeeded() {
+    // Set up environment.
+    let mut ledger = InMemoryLedger::with_bootstrap();
+    // Set up environment.
+    let token_by_recipient : Decimal =  Decimal::from(100); 
+    let mut test_env = TestEnv::new(&mut ledger, RADIX_TOKEN);
+    assert_eq!(test_env.get_balance(test_env.admin_account, RADIX_TOKEN).unwrap(), Decimal::from_str("1000000").unwrap());
```
cbisaillon: Instead of `Decimal::from_str().unwrap`, you can use `Decimal::from(1000000)` instead

basic/airdrop-with-withdraw/src/lib.rs
cbisaillon: I am not sure if this is what is causing the issue, but when you call `add_recipient`, the key you insert into the `recipients` map is the address of the account. Then in the `withdraw_token`, you try to find the entry in the map by specifying the address of the badge as key which does not exists.

I just realised that you will need to create a non fungible token instead of a badge for user identification, since every badges is the same so every user will be withdrawing from the same vault. Each NFT have a specific ID which can be used as the key in the `recipients` map.

cbisaillon (changes requested):

basic/airdrop-with-withdraw/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+            let admin_badge = ResourceBuilder::new_fungible(DIVISIBILITY_NONE)
+                                .initial_supply_fungible(1);
+
+            let minter_badge = ResourceBuilder::new_fungible(DIVISIBILITY_NONE)
+                                               .metadata("name", "minter badge")
+                                               .initial_supply_fungible(1); 
+
+            let recipient_badge_def = ResourceBuilder::new_fungible(DIVISIBILITY_NONE)
+                                .flags(MINTABLE | INDIVIDUAL_METADATA_MUTABLE)
+                                .badge(minter_badge.resource_address(), MAY_MINT | MAY_TRANSFER | MAY_CHANGE_INDIVIDUAL_METADATA)
```
cbisaillon, line 26: Here you don't need the `MAY_TRANSFER` and `MAY_CHANGE_INDIVIDUAL_METADATA` authorizations. 

`MAY_TRANSFER` is only used when the flag `RESTRICTED_TRANSFER` is set.
`MAY_CHANGE_INDIVIDUAL_METADATA` and `INDIVIDUAL_METADATA_MUTABLE` are used for NFTs only, when you need to call `ResourceDef.update_nft_data()`.

Take a look at this page for more info on flags and authorization: https://docs.radixdlt.com/main/scrypto/scrypto-lang/resources.html

cbisaillon (changes requested):

basic/airdrop-with-withdraw/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+            self.recipients.insert(recipient, Vault::with_bucket(tokens));
+            let user_badge = self.minter_badge_vault.authorize(|auth|
+                return self.recipient_badge_def.mint(1, auth) 
+            );
+            Account::from(recipient).deposit(user_badge);
+        }
+        
+
+        #[auth(recipient_badge_def)]
+        pub fn withdraw_token(&mut self) -> Bucket
```
cbisaillon, line 54: I am not sure if this is what is causing the issue, but when you call add_recipient, the key you insert into the recipients map is the address of the account. Then in the withdraw_token, you try to find the entry in the map by specifying the address of the badge as key which does not exists.

I just realised that you will need to create a non fungible token instead of a badge for user identification, since every badges is the same so every user will be withdrawing from the same vault. Each NFT have a specific ID which can be used as the key in the recipients map.

I suggest looking at my example on how to create NFTs: https://github.com/cbisaillon/Scrypto-Advent-Calendar/blob/master/day17/src/lib.rs. Notice how the ID of the NFT is just the number of NFT created. That way you are certain that it will always be unique

yetininoliviercoulibaly: Hello Clement, 
Thanks for the review. Indeed, I used account  address  as a key when insert into HashMap and when retrieving the address of the badge. I just updated the implementation with non fungible token  but I still have the Insufficient balance problem.

basic/airdrop-with-withdraw/tests/lib.rs
yetininoliviercoulibaly: done

basic/airdrop-with-withdraw/src/lib.rs
yetininoliviercoulibaly: done

basic/airdrop-with-withdraw/src/lib.rs
cbisaillon: The problem is the line `let recipient_badge = receipt.resource_def(1).unwrap(); `. This references the minter badge, not the recipient's badge. The index should be 2.

basic/airdrop-with-withdraw/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+        {  
+            let recipient_badge_id  = auth.get_nft_id();  
+            assert!(self.recipient_address_by_badge_id.contains_key(&recipient_badge_id), "Recipient not found");
+            let mut nft_data : AirdropWithWithdrawData = self.recipient_badge_def.get_nft_data(recipient_badge_id);
+            nft_data.is_collected = true;
```
cbisaillon, line 83: A user can withdraw multiple times since you are not making sure that `nft_data.is_collected == false`. Another way would be to burn the NFT

basic/airdrop-with-withdraw/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+    is_collected : bool
+}
+
+blueprint! {
+
+    
+    struct AirdropWithWithdraw {
+        admin_badge: Address,
+        tokens : Vault,
+        recipient_address_by_badge_id : HashMap<u128, Address>,
```
cbisaillon, line 17: This does not have to be a HashMap, you are only using it to check if a NFT id is present in the keys. You could use a `HashSet<u128>` instead

basic/airdrop-with-withdraw/src/lib.rs
yetininoliviercoulibaly: Thanks for the review, I just updated the implementation and it work fine now.

basic/airdrop-with-withdraw/src/lib.rs
yetininoliviercoulibaly: Indeed, I added the assert to check nft_data.is_collected value.

basic/airdrop-with-withdraw/src/lib.rs
yetininoliviercoulibaly: I removed the hasmap  because checking of  nft_data.is_collected is enough

cbisaillon:
Thank you for the example and good job !

cbisaillon (approved):

PR #53, opened 2021-12-30
First pass at a UBI token in Scrypto.

michaelwoodson:
A UBI token that allocates 1 token per person per epoch. Ideally there would be a 20% burn on each transaction but currently the burn is optional. Will enforce when `RECALLABLE` support is added.

It was fun to learn and refreshing to get introduced to Scrypto!

### DIFF
--- a/basic/ubi/Cargo.toml
+++ b/basic/ubi/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "ubi"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+sbor = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v0.2.0" }
+scrypto = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v0.2.0" }
+
+[dev-dependencies]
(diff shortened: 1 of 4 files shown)


cbisaillon (changes requested):
Thanks for this example. Here are some changes that you would need to do before I can merge on main. Can you also add more comments to help explain your code ?

cbisaillon:
@michaelwoodson I suggest joining the Scrypto mailing list to be eligible to 75$ in XRD when I merge this PR, if you have not already did: https://developers.radixdlt.com/devprogram

michaelwoodson:
@cbisaillon Thanks for the detailed feedback! Just pushed a change that should incorporate all the requested changes. I also signed up for the mailing list a couple weeks ago.

cbisaillon:
Good job. Thank you for doing the requested changes !

cbisaillon (approved):

cbisaillon:
You should receive your 75$ worth of XRD in the next few of weeks

michaelwoodson:
Very good, thank you!

PR #54, opened 2022-01-05
No loss lottery

BlindDev:
In this example, the lottery will use staking rewards as a winning pool

Players will be able to withdraw the amount they used to purchase tickets once the lottery is ended.


other/no-loss-lottery/staking/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+        staking_pool: Vault,
+        rewards_pool: Vault,
+        // reward value by staking pool amount
+        reward_value: Decimal,
+        // last update epoch
+        last_update: u64,
+        // reward rate per staking
+        rate: Decimal,
+        // collects rewards per account
+        rewards: HashMap<Address, Decimal>,
```
cbisaillon, line 15: Now that we have NFTs, you could create a NFT instead of a badge when users register and store the rewards, balances and paid amount inside its metadata.

other/no-loss-lottery/staking/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// Registers a new user
+        pub fn new_user(&self) -> Bucket {
+            ResourceBuilder::new_fungible(DIVISIBILITY_NONE)
+                .metadata("name", "Staking User Badge")
+                .initial_supply_fungible(1)
+        }
+
+        // staking for specific account
+        pub fn stake(&mut self, user_auth: BucketRef, bucket: Bucket){
```
cbisaillon, line 120: Here, instead of having the user_auth in the params, you could specify it with `#[auth(user_nft_resource_def)]`. Assuming you did the changes to use NFTs instead.

other/no-loss-lottery/lottery/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+        // address of staking contract
+        staking_address: Address,
+        // staking user badge
+        staking_user: Vault,
+        // vault that will be used for withdrawals
+        withdraw_vault: Vault,
+        // resource def for future minting
+        tickets: ResourceDef,
+        // players staking amount
+        players: HashMap<Address, HashMap<String, Decimal>>,
```
cbisaillon, line 130: Having maps with account addresses as key is an anti-pattern we are trying to avoid. You could instead store the `HashMap<String, Decimal>` inside the player's ticket metadata

other/no-loss-lottery/lottery/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+                debug!("Reward: {}", reward.amount());
+                self.ended.insert(self.lottery_name.to_string(), (winner, Vault::with_bucket(reward)));
+            });
+                        
+            // clear minted array
+            self.minted.clear();
+            debug!("=======================");
+        }
+
+        pub fn buy_ticket(&mut self, player: Address, payment: Bucket) -> (Bucket, Bucket) {
```
cbisaillon, line 244: Storing the stake data inside the NFT would allow you to simplify your code. You would not need to require a player_address in the params of this method

cbisaillon (changes requested):
Thanks for this example. I have left some comments. Can you also add more comments explaining what the methods are supposed to be doing ?

BlindDev:
Hey @cbisaillon, thank you so much for your feedback. Will work on updates!

other/no-loss-lottery/staking/src/lib.rs
BlindDev: Hey @cbisaillon. I'm not sure I understand the solution. Let's say I made this for registering a user:

```
pub fn new_user(&self) -> Bucket {
            // mint course NFT
            let token_data = StakingData {
                reward: Decimal::zero(),
                balance: Decimal::zero(),
                paid: Decimal::zero()
            };

            let bucket = self.token_minter.authorize(|auth| {
                self.token_resource_def
                    .mint_nft(self.token_id_counter, token_data, auth)
            });
            self.token_id_counter += 1;
            
            bucket
        }
``` 

So now the user has an NFT token, then how's the #auth will work if I need to update the user's NFT metadata?
I thought that I should put this NFT in addition to the staking amount so its metadata will be updated and the NFT will be returned back to the user after the update. Is the following code a valid pattern?

 ```
pub fn stake(&mut self, token: Bucket, staking: Bucket) -> Bucket {

             // get nft data from token
            let id = token.get_nft_id();
            let mut data: StakingData = token.get_nft_data(id);

            let reward = self.update_reward(data);
            let amount = staking.amount();

            data.reward = reward;
            data.paid = self.reward_value;
            data.balance += amount;

            self.token_minter
            .authorize(|auth| token.update_nft_data(id, data, auth));

            debug!("Account staking balance: {}", data.balance);

            self.staking_pool.put(staking);

            token
        }
```

cbisaillon: Then, instead of doing `token.update_nft_data`, you would use the resource_def: `self.nft_resource_def.update_nft_data()`

BlindDev: Wow, this is cool. Working on the update, thank you!

BlindDev:
@cbisaillon please check the code I've pushed. Right now there will be three different NFTs: 
Staking token (controls staking data, used as auth), 
Lottery token (controls lottery progress, winners, minted tickets)
Lottery ticket (used as auth for the ticket holder for withdrawal)

cbisaillon:
Thank you very much for doing the requested changes. This is a really cool example, good job !

PR #56, opened 2022-01-17
feat: DeFi Lending Platform

bdebenon:


defi/lending_platform/setup_and_run.sh
```diff
@@ -0,0 +1,90 @@
+resim reset
```
cbisaillon, line 1: I suggest looking into revup (https://github.com/RadGuild/revup). It allows you to do things like that but in a more readable way.

defi/lending_platform/src/lending_platform.rs
```diff
@@ -0,0 +1,436 @@
...
+                    info!(
+                        "[LendingPlatform][POOL] Added asset {} with LTV ratio of {} to the lending pool.",
+                        asset_address,
+                        ltv_ratio
+                    )
+                }
+            };
+        }
+
+        pub fn deposit_asset(&self, user_auth: BucketRef, asset: Bucket) -> BucketRef {
```
cbisaillon, line 156: No need to return the BucketRef here. You can just call `user_auth.drop()` somewhere inside the method. Or better: Make the user auth a NFT, use the `#[auth(user_auth_nft)]` macro which will automatically drop the reference

defi/lending_platform/src/lending_platform.rs
```diff
@@ -0,0 +1,436 @@
...
+            asset_address: Address,
+            amount: Decimal
+        ) -> (BucketRef, Bucket) {
+            let mut user = self.users.get(&user_auth.resource_address()).unwrap();
+            let user_address = user.user_address;
+
+            let removed_asset = match self.assets.get(&asset_address) {
+                Some(x) =>  {
+                    assert! (
+                        x.amount() >= amount,
```
cbisaillon, line 201: Shouldn't you take the `asset_ltv_ratios` into account here to make sure the user is over-collateralizing their loan ?

defi/lending_platform/src/lending_platform.rs
```diff
@@ -0,0 +1,436 @@
...
+            asset_address: Address,
+            amount: Decimal
+        ) -> (BucketRef, Bucket) {
+            let mut user = self.users.get(&user_auth.resource_address()).unwrap();
+            let user_address = user.user_address;
+
+            let borrowed_asset = match self.assets.get(&asset_address) {
+                Some(x) =>  {
+                    assert! (
+                        x.amount() >= amount,
```
cbisaillon, line 249: asset_ltv_ratios

cbisaillon, line 249: Should you take into account the `asset_ltv_ratios` here to make sure the user is over-collateralizing their loan ?

defi/lending_platform/src/lending_platform.rs
```diff
@@ -0,0 +1,436 @@
...
+                "[LendingPlatform][USER:{}] Repaying {} of {}.",
+                user_address,
+                amount,
+                asset_address
+            );
+
+            match self.assets.get(&asset_address) {
+                Some(x) =>  {
+                    x.put(asset);
+                    let updated_asset_amount = self.assets.get(&asset_address).unwrap().amount();
```
cbisaillon, line 322: I think you can just use `x.amount()` here

defi/lending_platform/src/lending_platform.rs
```diff
@@ -0,0 +1,436 @@
...
+    struct LendingPlatform {
+        assets: LazyMap<Address, Vault>,
+        asset_ltv_ratios: LazyMap<Address, Decimal>,
+        loan_balances: LazyMap<Address, Decimal>,
+        users: LazyMap<Address, User>,
+        admin_badge: ResourceDef
+    }
+
+    impl LendingPlatform {
+        fn calculate_total_collateral(&self, user: &User) -> Decimal {
```
cbisaillon, line 14: Can't you keep track of the `total_collateral` and `total_loan_balance` inside variables on the User struct ? That way you don't have to loop each time

bdebenon: I don't believe so. The reason is that in the future we will be using a price oracle that will provide the current prices of each asset which will need to be taken into account each time we need access the `total_collateral` or `total_loan_balance`.

Additionally, if we were to compute theses values on the User level, would would have to pass asset prices or an oracle into the User object which would in opinion break the domain boundaries of a `User`

defi/lending_platform/src/lending_platform.rs
bdebenon: Nice catch! Updated

defi/lending_platform/src/lending_platform.rs
```diff
@@ -0,0 +1,436 @@
...
+
+            // Check if user has enough collateral available for the loan
+            let user_available_collateral = self.calculate_available_collateral(&user); // TODO: Continue here
+            info!(
+                "[LendingPlatform][USER:{}] Available collateral in terms of XRD: {}",
+                user_address,
+                user_available_collateral
+            );
+            assert! (
+                user_available_collateral >= borrow_amount_in_terms_of_xrd,
```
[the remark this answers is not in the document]
bdebenon: The function `calculate_total_collateral` that gets used when computing `user_available_collateral` takes into account LTV. Adding a comment to make that more clear!

defi/lending_platform/setup_and_run.sh
bdebenon: Looks interesting. Will take a look in the near future!

defi/lending_platform/src/lending_platform.rs
bdebenon: Updated it to use `user_auth.drop()`.

To use the `#[auth(user_auth_nft)]` macro, would I need to store information about the badge within the component similar to how I did the `admin_badge` auth? Or is there an alternative way to do this that doesn't require having access to `user_auth_nft` from within the component?

cbisaillon: You would do exactly the same thing as you are doing right now but instead of using the `user_auth.resource_address` as keys to the maps, you would use the NFT id

bdebenon: What would the function signature look like? If I do
```
#[auth(user_auth)]
pub fn deposit_asset(&self, asset: Bucket)
```

I get the following error because the nft does not belong to the component
```
error[E0425]: cannot find value `user_auth` in this scope
   --> src/lending_platform.rs:167:44
    |
167 |             let mut user = self.users.get(&user_auth.resource_address()).unwrap();
    |                                            ^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `user_auth` in this scope
   --> src/lending_platform.rs:180:31
    |
180 |             self.users.insert(user_auth.resource_address(), user);
    |                               ^^^^^^^^^ not found in this scope

error[E0609]: no field `user_auth` on type `&lending_platform::blueprint::LendingPlatform`
   --> src/lending_platform.rs:162:16
    |
162 |         #[auth(user_auth)]
    |                ^^^^^^^^^ unknown field
    |
    = note: available fields are: `assets`, `asset_ltv_ratios`, `loan_balances`, `users`, `admin_badge`
```

Not sure how to achieve this without making the badges part of the struct

cbisaillon: To reference the resource used as authentication, you use the `auth` variable. In that case, you used the name of the variable (`user_auth`)

cbisaillon: Let's say you name the NFT resourceDef `auth_nft` in you component's state:
```
struct LendingPlatform {
   assets: LazyMap<Address, Vault>,
   asset_ltv_ratios: LazyMap<Address, Decimal>,
   ...
   auth_nft: ResourceDef 
}
```

You would reference it like this in the auth macro:

```
#[auth(auth_nft)]
pub fn deposit_asset(&self, asset: Bucket)
```

The problem might be that you are not saving the NFT resource def in the state ?

cbisaillon:
Sorry if it took so long for me to reply. 
Could you update this example to use Scrypto v0.3 please ?

0xOmarA:
I'm closing this PR since there has not been much activity here for a while. Would be happy to reopen it if @bdebenon decides to continue working on this example.

PR #58, opened 2022-02-07
Zombo Bakery

MisoShiru69:


other/zombo-bakery/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+        //Need bread to purchase butter
+        #[auth(bread)]
+        pub fn buy_butter(&self,payment:Bucket)->(Bucket,Bucket){
+            info!("{}", "Bread and butter is the best!");
+            (self.material_minter.authorize(|auth| self.butter.mint(1, auth)),payment)
+        }
+
+        //Enjoy your bread
+        //Need bread butter and NFT to eat bread
+        pub fn eat_bread(&self, bread:Bucket, butter:Bucket, nft_bucket:Bucket)->Bucket{
```
cbisaillon, line 192: For the `nft_bucket`, you could request a BucketRef instead of a Bucket. That way, you assure the user that you are not going to steal their badge and store it in your vault. Also, you will not have to return the bucket back. You only have to drop it at the end: `nft_bucket.drop()`

other/zombo-bakery/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+            self.material_minter.authorize(|auth| self.bread.burn_with_auth(bread, auth));
+            self.material_minter.authorize(|auth| self.butter.burn_with_auth(butter, auth));
+            info!("{}", "Damn, that's some good bread!");
+
+            nft_bucket
+
+        }
+        //Now that you are sick of bread, lets make somthing else
+        //Need NFT weight >= 3 
+        pub fn next_level(&self, nft_bucket:Bucket) ->  (Bucket, Bucket) {
```
cbisaillon, line 216: You could also request a BucketRef here instead of a Bucket

cbisaillon (changes requested):
Really cool example ! Good job

other/zombo-bakery/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+                baker_nft_minter:Vault::with_bucket(nft_minter),
+                baker_nft_def,
+                baker_nft_price:50.into(),
+                baker_card_id_counter:0,
+                bread:bread_def,
+                butter:butter_def,
+                cake_pan:cake_pan_def,
+                collected_xrd:Vault::new(RADIX_TOKEN),
+                flour:flour_def,
+                material_price:100.into(),
```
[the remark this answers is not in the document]
MisoShiru69: Added method to change prices from the default values.

PR #61, opened 2022-02-14
Created a Payment Splitter example

0xOmarA:
Hi! 

I've created a payment splitter example using Scrypto which is inspired by the OpenZeppelin [implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/finance/PaymentSplitter.sol) of a Payment Splitter.

This PaymentSplitter uses NFTs to authenticate shareholders and to keep track of the amount of shares that a shareholder owns, the amount of XRD the payment splitter owes them in total, and the amount that has been given so far.

I've added an `example.py` file to run this example quickly using python and resim and have also included a detailed example in the `README.md` file.


basic/payment-splitter/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "payment_splitter"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+sbor = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v0.2.0" }
```
cbisaillon, line 7: Please update to v0.3.0

basic/payment-splitter/README.md
```diff
@@ -0,0 +1,173 @@
...
+* The script will then add all four of the addresses that we created as shareholders by calling the `add_shareholder` method on the PaymentSplitter component.
+* Some XRD will be deposited by account 1 (by calling the `deposit_xrd` method) and then later account 2 will attempt to withdraw their share of it (by calling the `withdraw_xrd` method).
+
+You do not need to install any additional packages to run this `example.py` file. This file used the standard python 3 library and packages. So, to run this example just type the following into your command line:
+
+```shell
+python3 example.py
+```
+
+### Method 2: Manual method using Resim and CLI.
```
cbisaillon, line 32: Would be cool to also have `Method 3: Using transaction manifest files` :)

basic/payment-splitter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
+use scrypto::prelude::*;
+
+/// Defines the data of the shareholders badge which is an NFT.
+#[derive(NftData)]
```
cbisaillon, line 4: Objects related to NFTs were renamed in v0.3. Please look at https://docs.radixdlt.com/main/scrypto/release_notes/v0_3.html#_changed_naming_of_nft_related_objects_and_methods

basic/payment-splitter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+                    shares: shareholder_shares,
+                    total_xrd_owed: Decimal(0),
+                    total_xrd_withdrawn: Decimal(0)
+                },
+                auth
+            );
+            self.number_of_shareholders += 1;
+
+            // Airdropping the NFT to the user with the specified address
+            Account::from(shareholder_address).deposit(nft_bucket);
```
cbisaillon, line 126: We removed the Account API with v0.3. There is an example here: https://docs.radixdlt.com/main/scrypto/release_notes/v0_3.html#_removed_the_account_api on sending funds to an account component. 

But, what we recommend, is not to send directly to a wallet. Instead, return the NFT and let the caller send the token to the address they want. Calling `add_shareholder` and depositing to the account can be achieved in a single transaction with a manifest file

basic/payment-splitter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+            // Getting the amount of XRD that passed in the XRD bucket
+            let xrd_amount: Decimal = xrd_bucket.amount();
+            info!("Depositing XRD of amount: {}", xrd_amount);
+
+            // Taking the XRD bucket passed and depositing it into the XRD vault
+            self.xrd_vault.put(xrd_bucket);
+
+            // Getting the total number of shares that we have across all of our shareholders
+            let mut total_quantity_of_shares: Decimal = Decimal(0);
+            for i in 0..self.number_of_shareholders {
```
cbisaillon, line 150: Instead of doing that loop every time XRD is added, you could keep a field `total_quantity_of_shares` on the component state and update it inside the `add_shareholder` method

basic/payment-splitter/tests/lib.rs
```diff
@@ -0,0 +1,36 @@
+use radix_engine::ledger::*;
```
cbisaillon, line 1: You can remove this file

basic/payment-splitter/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+            .metadata("symbol", "iadm")
+            .initial_supply_fungible(1);
+
+            // Creating the shareholders NFT Resource Definition. We will not mint any of them at this point
+            // of time as the minting and creation of these tokens happens when the `add_shareholder` function
+            // is called
+            let shareholder_def: ResourceDef = ResourceBuilder::new_non_fungible()
+                .metadata("name", "Shareholder")
+                .metadata("symbol", "shr")
+                .flags(MINTABLE | INDIVIDUAL_METADATA_MUTABLE | SHARED_METADATA_MUTABLE)
```
cbisaillon, line 78: You don't need the flag `SHARED_METADATA_MUTABLE` and the permission `MAY_CHANGE_SHARED_METADATA` since you are only mutating **individual** metadata in your example

cbisaillon (changes requested):

0xOmarA:
Thank you so much Clement for the super detailed comments. I've made the changes accordingly and have fully ported this code over to use Scrypto v0.3.0. 

In addition to that I've added the transaction manifest files for the tests which was a lot of fun to learn and to work on!

cbisaillon:
Thank you for doing the update. When I run `python3 example.py`, I get the following error:

```
Traceback (most recent call last):
  File "example.py", line 91, in <module>
    main()
  File "example.py", line 59, in main
    package: str = re.findall(r'Package: (\w+)', response)[0]
IndexError: list index out of range
```

0xOmarA:
I have made a number of modifications to the code and to the documentation. Here are a few notes as to what I've done:

* Now each shareholder in the `PaymentSplitter` has their own vault. I believe that this to be a better design from an asset-oriented perspective and that this could reduce the risks of XRD getting mixed up.
* I've removed the `rtm_creator.py` script and instead added instructions the `README.md` file to help the readers use the already created`.rtm` files with the same addresses as the ones I had on my local machine. 
* Made a number of minor modifications to the `README.md` file to reflect on the change made to the blueprint whereby it no longer requires an `internal_admin_badge`.

basic/payment-splitter/script.sh
```diff
@@ -0,0 +1,20 @@
+resim reset
+
+export op1=$(resim new-account)
```
cbisaillon, line 3: I am getting `error: Found argument 'account' which wasn't expected, or isn't valid in this context` when running this script

basic/payment-splitter/src/lib.rs
```diff
@@ -150,37 +128,37 @@ blueprint! {
...
-                self.internal_admin_badge.authorize(|auth| {
-                    self.shareholder_def.update_non_fungible_data(&key, shareholder, auth);
-                })
+                let nft_id: NonFungibleKey = NonFungibleKey::from(i);
+                
+                // Adding the amount of XRD owed to this shareholder to their respective vault
+                let shareholder: ShareHolder = self.shareholder_def.get_non_fungible_data(&nft_id);
+                let xrd_owed_to_shareholder: Decimal = xrd_amount * shareholder.shares / self.total_quantity_of_shares;
+
+                // self.xrd_vaults[nft_id].put(xrd_bucket.take(xrd_owed_to_shareholder));
```
cbisaillon, line 157: You can remove this comment now

basic/payment-splitter/script.sh
0xOmarA: Apologies, `script.sh` was a file that I was experimenting with internally and did not mean to commit it. I've removed this file in the most recent commit as I did have similar issues with it as well.

PR #63, opened 2022-03-24
Add an example for an order book based dex

backyard-coder:


defi/order-book-dex/src/trading_pair.rs
```diff
@@ -0,0 +1,311 @@
...
+                .metadata(
+                    "name",
+                    format!(
+                        "Dex limit order ({}/{})",
+                        base_resource.address(),
+                        quote_resource.address()
+                    ),
+                )
+                .flags(MINTABLE | BURNABLE | INDIVIDUAL_METADATA_MUTABLE)
+                .badge(minter.resource_def(), ALL_PERMISSIONS)
```
cbisaillon, line 50: It's ok right now to put `ALL_PERMISSIONS`, but you wouldn't want that for production because it would mean that the minter authority would be able to update the flags (MAY_MANAGE_RESOURCE_FLAGS). It's not a big deal since the badge is stored on the component and you don't provide methods to change the resource flags, but we are planning on having the wallet display warnings when a token might be recallable or restricted in the future

cbisaillon (approved):
This is a really high quality example!

defi/order-book-dex/src/trading_pair.rs
backyard-coder: Hmm, I totally missed the MAY_MANAGE_RESOURCE_FLAGS flag and thought the ALL_PERMISSIONS flag only pertained to the set flags (MINTABLE, BURNABLE, INDIVIDUAL_METADATA_MUTABLE). I will correct that and only use the necessary permissions.

Now I'm a bit sceptical about the ALL_FLAGS and ALL_PERMISSIONS constants in general. In security centric code like blueprints, would we not be well adviced to only very selectively set flags and permissions? Usage of these constants should be discouraged imho.

cbisaillon: I'll ask to the team about that

PR #75, opened 2022-04-20
Guess it game

TechNinjaWeb:
[WIP] Very basic dice roll guessing game. 

- Create a game with a minimum bet amount
- Only 2 players may enter a game
- It's single round at the moment, but can be extended to multiple rounds
- Players will make a guess and a dice is rolled
- The player closest to the guess without going over is winner
- Winner with the correct NFT token ID can withdraw funds

Looking for any feedback on how to improve the readability, optimize the code, and or determine a better strategy for things like RNG and game state transitions.

There is a fake frontend associated with this repo that just loads up an express server and forwards shell commands from REST endpoints.


games/guess-it/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "guess-it"
+version = "0.1.0"
+edition = "2021"
+description = "A simple guessing game. 2 round submission. 1st round is a hash of the answer, 2nd is to reveal the secret."
```
TechNinjaWeb, line 5: changed to single round. will update on next commit

games/guess-it/README.md
```diff
@@ -0,0 +1,34 @@
+## Scrypto Unit: A lightweight testing framework for Scrypto
```
TechNinjaWeb, line 1: this file is somewhat unnecessary, but it was helpful in helping me get scrypto-unit up and running properly

games/guess-it/server/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "server",
+  "version": "0.0.0",
+  "private": true,
+  "scripts": {
+    "start": "nodemon ./bin/www",
+    "start:dev": "DEBUG=server:* npm start"
```
TechNinjaWeb, line 7: load up http://localhost:3000/radix

games/guess-it/server/routes/index.js
```diff
@@ -0,0 +1,9 @@
+var express = require('express');
+var router = express.Router();
+
+/* GET home page. */
+router.get('/', function(req, res, next) {
```
TechNinjaWeb, line 5: Useless route. Will replace with default Radix route

games/guess-it/server/routes/radix.js
```diff
@@ -0,0 +1,286 @@
...
+router.post('/join-game', async function(req, res, next) {
+    const { public, private, component, bet } = req.body;
+    const response = await join_game(public, private, component, bet);
+    if (response.error){ return res.status(500).json(response) }
+
+    return res.json(response);
+});
+
+router.post('/make-guess', async function(req, res, next) {
+    const { public, private, component, guess } = req.body;
```
TechNinjaWeb, line 127: Likely the first 2 params could go since they are being stored on the server during io.connection phase

games/guess-it/server/routes/users.js
```diff
@@ -0,0 +1,9 @@
+var express = require('express');
```
TechNinjaWeb, line 1: Boilerplate route. Will be removed next comment

games/guess-it/server/views/index.ejs
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
TechNinjaWeb, line 1: Boilerplate. Can be removed next commit

games/guess-it/src/context.rs
```diff
@@ -0,0 +1,17 @@
+use scrypto::prelude::*;
```
TechNinjaWeb, line 1: Unused atm, but was using to figure out how to `set_epoch` after I've passed the `mut &ledger` into the `TestEnv`. 

Need more help here!

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+    type Err = String;
+
+    fn from_str(input: &str) -> Result<State, Self::Err> {
+        match input {
+            "AcceptingChallenger" => Ok(State::AcceptingChallenger),
+            "MakeGuess" => Ok(State::MakeGuess),
+            "SubmitSecret" => Ok(State::SubmitSecret),
+            "WinnerSelection" => Ok(State::WinnerSelection),
+            "Payout" => Ok(State::Payout),
+            "Refunding" => Ok(State::Refunding),
```
TechNinjaWeb, line 32: Didn't implement!

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+    Destroyed = 6,
+}
+impl FromStr for State {
+    type Err = String;
+
+    fn from_str(input: &str) -> Result<State, Self::Err> {
+        match input {
+            "AcceptingChallenger" => Ok(State::AcceptingChallenger),
+            "MakeGuess" => Ok(State::MakeGuess),
+            "SubmitSecret" => Ok(State::SubmitSecret),
```
TechNinjaWeb, line 29: Didn't implement! Moved from 2 rounds to a single round and a secret is no longer needed

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn from(input: u32) -> State {
+        match input {
+            0 => State::AcceptingChallenger,
+            1 => State::MakeGuess,
+            2 => State::SubmitSecret,
+            3 => State::WinnerSelection,
+            4 => State::Payout,
+            5 => State::Refunding,
+            6 => State::Destroyed,
+            _ => State::Refunding,
```
TechNinjaWeb, line 49: Cleanup states on next commit

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            last_roll: game.last_roll.clone(),
+        };
+        unescape(serde_json::to_string(&serializable).unwrap_or("".to_string())
+            .as_mut_str())
+            .unwrap_or("".to_string())
+    }
+}
+
+#[derive(TypeId, Decode, Encode, Clone, Deserialize, Serialize)]
+pub struct GameSerialized {
```
TechNinjaWeb, line 114: Maybe there's a better way to serialize the Game object?

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+    guess: String,
+}
+impl Player {
+    pub fn empty() -> Player {
+        return Self {
+            guess: "".into(),
+        }
+    }
+
+    pub fn get_guess(&self) -> String { self.guess.clone() }
```
TechNinjaWeb, line 152: likely unnecessary to have getter and setter. Was having issues and this method is a remnant.

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+    let multiplier = game.players.clone().into_iter()
+        .map(|(_,p)| u128::from_str(p.guess.as_str()).unwrap_or(0))
+        .reduce(|a,b| a * b).unwrap_or(1);
+
+    Uuid::generate() * multiplier
+}
+
+fn get_winner(game: &Game, random_number: u128) -> String {
+    let player_ids: Vec<NonFungibleKey> = game.players.clone().into_iter().map(|(k,_)| k).collect();
+    let p1 = random_number - u128::from_str(game.players.get(&player_ids[0]).unwrap().guess.as_str()).unwrap_or(0);
```
TechNinjaWeb, line 282: there's likely a better way to determine closest to the number.

games/guess-it/server/views/index.ejs
TechNinjaWeb, line 1: Boilerplate. Can be removed next commit

games/guess-it/README.md
0xOmarA: Could you change the contents of the `README.md` so that it includes some instructions on how people can play the game? Like what sort of functions and methods need to be called in what order and perhaps an example.

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            }.instantiate()
+        }
+
+        pub fn join(&mut self, mut payment: Bucket) -> (Bucket, Bucket, String) {
+            assert_eq!(self.game.state, State::AcceptingChallenger, "Not accepting challengers!");
+            assert!(self.game.players.len() <= 2, "Too many players!");
+
+            let badge = self.badges.take(1.to_string());
+            let player = Player::empty();
+            let auth = badge.present();
```
0xOmarA, line 201: I'm assuming here that you're presenting a badge to get the `NonFungibleKey`. There if no need for you to do that. Instead, why not replace lines 199 ~ 203 with:
```rust
let player: Player = Player::empty();
let badge: Bucket = self.badges.take(1u32);
let nft_id: NonFungibleKey = badge.get_non_fungible_key();
```

Calling `present` on a `Bucket` creates a `BucketRef` which you can think of as an immutable read-only `Bucket`. By this logic, you should be able to perform all of the read operations you need from the bucket directly.

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            let nft_id = auth.get_non_fungible_key().clone();
+            auth.drop();
+            // Secure the funds
+            self.bank.put(payment.take(Decimal::from(self.game.bet.to_string())));
```
0xOmarA, line 205: There is no need to call `self.game.bet.to_string()`. What you want to use is `self.game.bet.clone()` to clone it and be able to use it in the function call.

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+impl Player {
+    pub fn empty() -> Player {
+        return Self {
+            guess: "".into(),
+        }
+    }
+
+    pub fn get_guess(&self) -> String { self.guess.clone() }
+    pub fn make_guess(&mut self, guess: String) { self.guess = guess; }
```
0xOmarA, line 153: The `guess` argument should not be a `String` here since this game is entirely based on numbers. A more appropriate type here would be `u8`, `u16`, `u32` or maybe `Decimal`. 

The use of a String means that somebody could put in a string which would later lead to undesired behavior happening.

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    Uuid::generate() * multiplier
+}
+
+fn get_winner(game: &Game, random_number: u128) -> String {
+    let player_ids: Vec<NonFungibleKey> = game.players.clone().into_iter().map(|(k,_)| k).collect();
+    let p1 = random_number - u128::from_str(game.players.get(&player_ids[0]).unwrap().guess.as_str()).unwrap_or(0);
+    let p2 = random_number - u128::from_str(game.players.get(&player_ids[1]).unwrap().guess.as_str()).unwrap_or(0);
+    let closest_guess = cmp::min(p1, p2);
+    let winner = if p1 == closest_guess { &player_ids[0] } else { &player_ids[1] };
```
0xOmarA, line 285: In the case where both players guess the same number, `p1` would have an advantage over `p2` as they would always win in the case of a tie.

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            else { panic!("You cannot withdraw funds!") }
+        }
+
+        pub fn get_context(&self) -> (Actor, Address, H256, u64, u128) {
+            context::ContextTest::query()
+        }
+    }
+}
+
+fn get_random(game: &Game) -> u128 {
```
0xOmarA, line 272: There are a few things about this `get_random` function:
1- This is a pseudorandom function which is okay for an example but not for a full working product. It's okay to keep it for now and there is no need to change it as this is just an example.

2- Your `multiplier` is always 0 which means that the `get_random` function always returns 0. What is the purpose of the `multiplier` in this case?

0xOmarA (changes requested):
Hey Raymond!

Happy to see that you've submitted your first example to the community scrypto repo. Overall, I think that this is a pretty good example with interesting functionality. However, I have a few comments on it:

1- The use of `String`s is a little bit heavy and sometimes they're used in places where they're inappropriate and where an unsigned number or `Decimal` could be better. Could you take a look at these few instances?

2- Methods and functions can return any data type. So feel free to return a `NonFungibleKey` when you need to instead of relying on strings.

games/guess-it/src/lib.rs
TechNinjaWeb: The multiplier should be just p1's guess and p2's guess added together and then multiplied by the modulus operation on the Uuid.

it was just a quick way to add some more randomness to the RNG.

games/guess-it/src/lib.rs
TechNinjaWeb: I'll address this in the next commit

games/guess-it/src/lib.rs
TechNinjaWeb: I was under the impression I'd tested this and thought only Strings could be interpreted from the resim command, but I'll change this going forward

games/guess-it/src/lib.rs
TechNinjaWeb: yep, thanks!

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+fn get_winner(game: &Game, random_number: u128) -> String {
+    let player_ids: Vec<NonFungibleKey> = game.players.clone().into_iter().map(|(k,_)| k).collect();
+    let p1 = random_number - u128::from_str(game.players.get(&player_ids[0]).unwrap().guess.as_str()).unwrap_or(0);
+    let p2 = random_number - u128::from_str(game.players.get(&player_ids[1]).unwrap().guess.as_str()).unwrap_or(0);
+    let closest_guess = cmp::min(p1, p2);
+    let winner = if p1 == closest_guess { &player_ids[0] } else { &player_ids[1] };
+
+    return winner.to_string();
```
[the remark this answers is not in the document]
TechNinjaWeb: Ideally I'll do this, but I had some issues with trying to serialize structs that contained NFKeys. This is the reason lots of strings were reluctantly used

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+#[derive(TypeId, Decode, Encode)]
+pub struct Game {
+    name: String,
+    state: State,
+    bet: String,
+    players: HashMap<NonFungibleKey, Player>,
+    winner: String,
```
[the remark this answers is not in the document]
TechNinjaWeb: I had issues serializing structs with NftKeys and Decimals, so Strings were reluctantly used

games/guess-it/src/context.rs
cbisaillon: There is a better way to do this.

```
let mut executor = TransactionExecutor::new(ledger, false);
executor.ledger_mut().set_epoch(10);
```

games/guess-it/src/lib.rs
cbisaillon: I think `guess` is the hash of the number the user is committing to bet on?

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+                bank: Vault::new(RADIX_TOKEN),
+                badges: Vault::with_bucket(player_badges),
+                badge_ref: player_badge_def,
+                game: Game::new(None, name, bet),
+            }.instantiate()
+        }
+
+        pub fn join(&mut self, mut payment: Bucket) -> (Bucket, Bucket, String) {
+            assert_eq!(self.game.state, State::AcceptingChallenger, "Not accepting challengers!");
+            assert!(self.game.players.len() <= 2, "Too many players!");
```
cbisaillon, line 197: Should be `self.game.players.len() < 2` ?

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+
+            let badge = self.badges.take(1.to_string());
```
cbisaillon, line 199: I suggest using `Decimal::one()`

games/guess-it/src/context.rs
TechNinjaWeb: Amazing! that's exactly what I was looking for!

games/guess-it/src/lib.rs
TechNinjaWeb: I am under the impression that <2 would only work for 0-1 players. 
the assert is attempting to limit to 2 players and error "Too many players!" if any more try to join

games/guess-it/src/lib.rs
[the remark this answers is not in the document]
TechNinjaWeb: this should be addressed next commit

games/guess-it/src/lib.rs
TechNinjaWeb: on second thought, this method was necessary for use in testing as the guess is locked behind private field

games/guess-it/src/lib.rs
TechNinjaWeb: To be on the safe side of this, I'll be adding a test to make sure the multiplier isn't 0
image

TechNinjaWeb: The major issue I didn't notice until adding tests was that the UUID could not be multiplied, as it overflows the u128 bounds. Instead of multiplying, dividing produced the desired effect of increasing the randomness of the roll.

If there's any other way to determine this without relying on oracles, I'm willing to implement it. 

image

games/guess-it/src/lib.rs
TechNinjaWeb: There was some confusion about the number of players admitted into the game, so the README.md was updated to reflect that only 2 players are allowed to enter.

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+impl GameSerialized {
+    pub fn empty(players: Option<HashMap<String, Player>>) -> Self {
+        Self {
+            name: "".to_string(),
+            state: State::AcceptingChallenger,
+            bet: 0u128,
+            players: players.unwrap_or(HashMap::new()),
+            winner: NonFungibleKey::from(0).to_string(),
```
beemdvp, line 159: Isnt this just `"0".to_string()`?

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+
+        pub fn join(&mut self, mut payment: Bucket) -> (Bucket, Bucket, String) {
+            assert_eq!(self.game.state, State::AcceptingChallenger, "Not accepting challengers!");
+            assert!(self.game.players.len() <= 2, "Too many players!");
+
+            let player: Player = Player::empty();
+            let badge: Bucket = self.badges.take(Decimal::one());
+            let nft_id: NonFungibleKey = badge.get_non_fungible_key();
+            // Secure the funds
+            self.bank.put(payment.take(Decimal::from(self.game.bet as i128)));
```
beemdvp, line 226: Maybe `Game.bet` should just be `i128`?

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+            // Add the player's details
+            self.game.players.insert(nft_id.clone(), player.clone());
```
beemdvp, line 228: Is `clone()` needed here?

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+            assert_eq!(self.game.state, State::Payout, "It's not time to get paid!");
+            let nft_id: NonFungibleKey = auth.get_non_fungible_key();
+
+            if nft_id == self.game.winner {
+                let amt = self.bank.amount();
+                let payout = self.bank.take(amt);
+
+                (payout, format!("You withdrew {:?} XRD!", amt))
+            }
+            else { panic!("You cannot withdraw funds!") }
```
beemdvp, line 280: Just a style preference but a simpler way to write is this failing fast:
```
assert!(nft_id == self.game.winner, "You cannot withdraw funds!");

let amt = self.bank.amount();
let payout = self.bank.take(amt);

(payout, format!("You withdrew {:?} XRD!", amt))
```
This way, you only deal with happy path without worrying about alternative flows to panic

games/guess-it/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+
+        #[auth(badge_ref)]
+        pub fn make_guess(&mut self, guess: u128) -> String {
+            assert_eq!(self.game.state, State::MakeGuess, "Not ready to make a guess!");
+
+            let key = auth.get_non_fungible_key();
+            let player = self.game.players.get_mut(&key).unwrap();
+            player.guess = guess;
+
+            self.game.state = if self.game.has_guessed() { State::WinnerSelection } else { State::MakeGuess };
```
beemdvp, line 246: Just style preference here, i really like using match:
```rust
self.game.state = match self.game.has_guessed() {
  true => State::WinnerSelection,
  false => State::MakeGuess 
};
```

Or maybe you can move this to the `Game`, namely `self.game.get_state()`

Very minor though

games/guess-it/src/lib.rs
TechNinjaWeb: On: 224 - 228: a move occurs and I get the error ` move occurs because `nft_id` has type `scrypto::prelude::NonFungibleKey`, which does not implement the `Copy` trait`

It seems I do not need the clone for the player tho.

Will update this in upcoming commit

games/guess-it/src/lib.rs
TechNinjaWeb: Thx. Will update next commit

games/guess-it/src/lib.rs
TechNinjaWeb: honestly was thinking of a way to use match.

Will update next commit

0xOmarA (commented):
Just had a chance to review the changes that you've made. I'm quite satisfied with the reduction in the use of the `String` type and with your current use of types.

games/guess-it/src/lib.rs
0xOmarA: I will review the changes you make in the next series of commits and then merge it.

TechNinjaWeb:
> Just had a chance to review the changes that you've made. I'm quite satisfied with the reduction in the use of the `String` type and with your current use of types.

Just making some last minute updates of things knocked off from the ToDo's list
- Handle a tie
- Messaging in frontend

cbisaillon:
@TechNinjaWeb could you email me at "clement@rdx.works" please?

PR #78, opened 2022-05-07
initial commit for Single Pool liquidity Example

atoumbre:
This is a simple example of a single liquidity pool blueprint. It uses AMM liquidity pool token macanic to ensure fair distribution of fees based on when liquidity provider deposit or remove tokens.


defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+
+            // Add tokens to the liquidity pool
+            self.pool.put(tokens.take(tokens.amount()));
+
+            // Return the LP tokens along with any remainder
+            (lp_tokens,tokens)
+        }
+
+        /// Collect fee for liquidity provider. will be added to the pool and LP/Token will be ajusted accordingly
+        pub fn collect_fee(&mut self,
```
0xOmarA, line 91: I do not believe that this method is doing what you think it is doing. I have a few questions here:

- The function name is a little confusing in comparison to what it does. The name is `collect_fee` but it seems to be adding liquidity to the liquidity pool without minting LP tokens. Is this intentional?
- The `tokens` bucket gets deposited into the `self.pool` vault and nothing gets taken out essentially meaning that that we have no buckets in this context. I'm unsure as to why the method returns `tokens` when it's empty. 

I could be a little bit confused about what this method does so feel free to provide explain it.

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+        /// along with any remainder.
+        pub fn add_liquidity(&mut self,
+             mut tokens: Bucket
+        ) -> (Bucket,Bucket) {
+
+            // Get the resource manager of the lp tokens
+            let lp_resource_manager = borrow_resource_manager!(self.lp_resource_address);
+
+            // Mint LP tokens according to the share the provider is contributing 
+            let mut supply_to_mint = tokens.amount() * (self.total_lp / self.total_liquidity);
```
0xOmarA, line 76: If we remove the entire liquidity from the pool and then try adding liquidity back again then `self.total_liquidity` would be zero which leads to an error. This would mean that once liquidity has been all removed, the pool would be rendered unusable and we can't add more liquidity to it.

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+        pub fn remove_liquidity(&mut self,
+             lp_tokens: Bucket
+        ) -> Bucket {
+            assert!(
+                self.lp_resource_address == lp_tokens.resource_address(),
+                "Wrong token type passed in"
+            );
+
+            // Withdraw the correct amounts of tokens A and B from reserves
+            let  mut to_remove = lp_tokens.amount()/(self.total_lp/self.total_liquidity);
```
0xOmarA, line 117: The math here does not seem correct here. Shouldn't this be: `lp_tokens.amount() * self.total_liquidity / self.total_lp`?

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            to_remove = to_remove.round(self.lp_divisibility,RoundingMode::TowardsNearestAndHalfTowardsZero); 
+
+            // Remain residual liquidity will be withdrawl on the last withdrawal  
+            let withdrawn = self.pool.take(std::cmp::min(to_remove, self.pool.amount()));
```
0xOmarA, line 121: I am a little bit uncertain here as to why `std::cmp::min(to_remove, self.pool.amount())` is required in this case. 

If we follow the equation that I provided in my previous comment: `to_remove = lp_tokens.amount() * self.total_liquidity / self.total_lp`. We can define the share in the pool as 

```
share = lp_tokens.amount() / self.total_lp
```

We can simplify the `to_remove` equation to be: 

```
to_remove = share * self.total_liquidity
```

In this case, the only way for to_remove to be larger than `self.total_liquidity` would be if `share` is larger than 1 which is impossible to happen as share can only have a value between 0 and 1 based on the equation that we defined for it above. Thus if `0 ≤ share ≤ 1`, then based on the relationship of `share` and `to_remove` we can say that `to_remove ≥ self.pool.amount()` holds true all of the time. Therefore, `std::cmp::min(to_remove, self.pool.amount())` would always return `to_remove` in all cases and under all circumstances.

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+
+            // Return the new LiquidityPool component, as well as the initial supply of LP tokens
+            (liquidity_pool,lp_tokens)
+        }
+
+        /// Adds liquidity to this pool and return the LP tokens representing pool shares
+        /// along with any remainder.
+        pub fn add_liquidity(&mut self,
+             mut tokens: Bucket
+        ) -> (Bucket,Bucket) {
```
0xOmarA, line 70: The other bucket returns should no longer be required one you make the change:

`self.pool.put(tokens.take(tokens.amount()))` => `self.pool.put(tokens)`

```suggestion
        ) -> Bucket {
```

0xOmarA (changes requested):
Hi atoumbre, thanks for submitting this example! There are a number of changes which I hope you can make to this example and I've highlighted them in the comments.

Also, can you please include a `README.md` file for this example explaining what is in this package, what does it intend to do, how it works, etc...?

defi/single-token-liquidity-pool-2/tests/lib.rs
```diff
@@ -0,0 +1,39 @@
+use radix_engine::ledger::*;
```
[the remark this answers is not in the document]
atoumbre: I will remove the test folder. test is done using zx scripts

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            lp_name: String,
+            lp_divisibility: u8,
+        ) -> (ComponentAddress,Bucket) {
+
+            // Check arguments
+            assert!(
+                !tokens.is_empty(),
+                "You must pass in an initial supply of token"
+            );
+
```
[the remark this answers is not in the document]
atoumbre: It's true, I may need help on how to do the check :)

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            let lp_tokens = ResourceBuilder::new_fungible()
+                .divisibility(lp_divisibility)
+                .metadata("symbol", lp_symbol)
+                .metadata("name", lp_name)
+                .mintable(rule!(require(lp_mint_badge.resource_address())), LOCKED)
+                .burnable(rule!(require(lp_mint_badge.resource_address())), LOCKED)
+                .initial_supply(tokens.amount());
+            let lp_resource_address = lp_tokens.resource_address();
+
+            let total_lp  = tokens.amount();
```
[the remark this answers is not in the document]
atoumbre: Fixed, It was a mistake

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    impl LiquidityPool {
+        /// Creates a LiquidityPool component and returns the component address
+        /// along with the initial LP tokens.
+        pub fn new(
+            tokens: Bucket,
+            lp_symbol: String,
+            lp_name: String,
+            lp_divisibility: u8,
```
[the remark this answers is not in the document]
atoumbre: I faced some  rounding issue in the liquidity withdraw process. but I can still use the highest maximum divisibility.

defi/single-token-liquidity-pool-2/src/lib.rs
atoumbre: This method should be call from other component to add "revenue" to the pool. Like fee collected after utilization of the liquidity available in the pool. Maybe "add_collected_fee" is less confusing ?

defi/single-token-liquidity-pool-2/src/lib.rs
atoumbre: Fixed by introducing lp_per_asset_ratio with initial value of 1 and reset to 1 if the pool is empty after a liquidity withdraw

defi/single-token-liquidity-pool-2/src/lib.rs
atoumbre: It's the same result (after testing). but I introduced lp_per_asset_ratio = lp_resource_manager.total_supply() / self.pool.amount() and = 1 if the pool is empty

defi/single-token-liquidity-pool-2/src/lib.rs
atoumbre: std::cmp::min(to_remove, self.pool.amount()) is used for rounding issues. Without it, the last withdraw fails du to difference introduced by roundings.

defi/single-token-liquidity-pool-2/src/lib.rs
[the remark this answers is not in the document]
0xOmarA: Sure thing! You can do by adding the following to your code.

```rust
assert_ne!(
    borrow_resource_manager!(tokens.resource_address()).resource_type(),
    ResourceType::NonFungible,
    "A liquidity pool can only have fungible tokens."
)
```

defi/single-token-liquidity-pool-2/src/lib.rs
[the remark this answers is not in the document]
0xOmarA: When the divisibility set to max you should face the least amount of rounding errors.

0xOmarA (changes requested):

defi/single-token-liquidity-pool-2/src/lib.rs
0xOmarA: The thing here is that there is no guarantee that anybody would call your component to pay the fee and that's the main issue here.

defi/single-token-liquidity-pool-2/src/lib.rs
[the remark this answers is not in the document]
atoumbre: Done ... Thanks for the tip

defi/single-token-liquidity-pool-2/src/lib.rs
atoumbre: This is not for people but for another components. for example. Imagine a staking pool : add_collected_fee can be call on regular basis to give reward to stakers.

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -18,14 +15,18 @@ blueprint! {
...
             lp_name: String,
-            lp_divisibility: u8,
         ) -> (ComponentAddress,Bucket) {
 
             // Check arguments
             assert!(
                 !tokens.is_empty(),
                 "You must pass in an initial supply of token"
             );
+            // Check arguments
```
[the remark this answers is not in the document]
atoumbre: fixed...

defi/single-token-liquidity-pool-2/src/lib.rs
```diff
@@ -66,42 +62,40 @@ blueprint! {
...
+        ) -> Bucket {
 
             // Get the resource manager of the lp tokens
             let lp_resource_manager = borrow_resource_manager!(self.lp_resource_address);
 
             // Mint LP tokens according to the share the provider is contributing 
-            let mut supply_to_mint = tokens.amount() * (self.total_lp / self.total_liquidity);
-            supply_to_mint = supply_to_mint.round(self.lp_divisibility,RoundingMode::TowardsNearestAndHalfTowardsZero);
+            let mut supply_to_mint = tokens.amount() * self.lp_per_asset_ratio;
+            supply_to_mint = supply_to_mint.round(DIVISIBILITY_MAXIMUM,RoundingMode::TowardsNearestAndHalfTowardsZero);
```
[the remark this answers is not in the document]
atoumbre: I have remove roudings but I still need to use std::cmp::min(to_remove, self.pool.amount()) to have la last withdraw to scucced.

0xOmarA (commented):
I have just reviewed the changes you've made to your code and I'm quite happy with them! Would you have time to write a short readme file for the blueprint? This is optional but it's preferred that you do :)

atoumbre:
RADME file just added

PR #79, opened 2022-05-12
RPG game community example

quwin:
Took a while to do, I hope reviewing it isn't a pain lol. I'm gonna continue working on it though, after I do a pull request for another smaller example.


games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+                .mint_non_fungible(&NonFungibleId::random(), chest_data));
+            let new_pants = self.system_vault.authorize(|| borrow_resource_manager!(self.item_nft)
+                .mint_non_fungible(&NonFungibleId::random(), pants_data));
+            let new_gloves = self.system_vault.authorize(|| borrow_resource_manager!(self.item_nft)
+                .mint_non_fungible(&NonFungibleId::random(), gloves_data));
+            let new_belt = self.system_vault.authorize(|| borrow_resource_manager!(self.item_nft)
+                .mint_non_fungible(&NonFungibleId::random(), belt_data));
+            let new_shoes = self.system_vault.authorize(|| borrow_resource_manager!(self.item_nft)
+                .mint_non_fungible(&NonFungibleId::random(), shoes_data));
+            let new_weapon = self.system_vault.authorize(|| borrow_resource_manager!(self.item_nft)
```
0xOmarA, line 339: In most cases, it is not recommended that you mint NFTs of multiple data types through the same resource. This makes your code be open to many different runtime errors due to types being incorrectly handled at runtime which is an issue. 

I recommend that you use three resources for your `Armor`, `Weapon`, and `Accessory `.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+                    return choice
+                }
+                else {
+                    continue
+                }
+            };
+        }
+        // Takes two of the same type + level Weapon/Armor/Accessory NFT, burns them, and makes a new one based off of the first. 
+        // Takes stats from first NFT. Stats increase by 20% per upgrade
+        pub fn fuse_items(&mut self, which: u8, item_bucket: Bucket) -> Bucket {
```
0xOmarA, line 479: I don't believe that the caller should pass the `which` argument to the method. Instead, I believe that the type of the NFT data can be inferred once you separate them so that one resource can't use multiple NFT data types.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+                assert!(item.id == item2.id);
+                assert!(item.part == item2.part);
+                item.level += dec!(1); 
+                item.health *= dec!("1.2");
+                item.defense *= dec!("1.2");
+                item.ability_scaling *= dec!("1.2");  
+                let new = Armor { id: item.id, part: item.part, level: item.level, health: item.health,
+                    defense: item.defense, weight: item.weight, ability: item.ability, ability_scaling: item.ability_scaling, version: self.version,
+                    };
+                self.system_vault.authorize(||item_bucket.burn());
```
0xOmarA, line 516: I see that this method uses `system_vault.authorize()` quite a number of times. Should this method be made to use `ComponentAuthZone::push(system_vault.create_proof())` at the top and `ComponentAuthZone::pop(system_vault.create_proof())` at the bottom instead?

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+                let new_bucket = self.system_vault.authorize(||
+                    borrow_resource_manager!(self.item_nft)
+                        .mint_non_fungible(&NonFungibleId::random(), new));
+                return new_bucket
+            }              
+        }
+        // Creates weapons
+        pub fn create_weapon_1(&mut self, gold: Bucket, resource1: Bucket, resource2: Bucket, class: u8) -> Bucket {
```
0xOmarA, line 544: I think that the `class` can be changed from a `u8` to be an enum to help make the code clearer.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+            assert!(gold.resource_address() == self.token_gold);
+            assert!(resource1.resource_address() == self.token_greavite);
+            assert!(resource2.resource_address() == self.token_wood);
+            let key_bucket: Bucket = self.system_vault.take(1);
+            let class_chosen = if class == 1 { String::from("Sword") }
```
0xOmarA, line 549: Why not use a `HashMap` for this which maps the class enum (which I commented about previously) to a tuple of values? Something like: `HashMap<ClassType, (String, Decimal, Decimal, Decimal, Decimal)>`.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+            self.system_vault.authorize(|| 
+                gold.burn());
+            self.system_vault.authorize(||
+                resource1.burn());
+            self.system_vault.authorize(||
+                resource2.burn());
+            self.system_vault.put(key_bucket);
+            new_weapon
+        }
+        pub fn create_armor_1(&mut self, gold: Bucket, resource1: Bucket, resource2: Bucket, part_chosen: u8) -> Bucket {
```
0xOmarA, line 610: The method name having `1` in there, implies that there could be `2` or `3`, but there is not. Does it need to have the `1` there?

games/rpg/src/lib.rs
0xOmarA, line 544: The method name having `1` in there, implies that there could be `2` or `3`, but there is not. Does it need to have the `1` there?

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+                ability: dec!(0),
+                ability_scaling: ability_scale,
+                range: dec!(1),
+                version: dec!(1),
+            };
+            let new_weapon = self.system_vault.authorize(||
+                borrow_resource_manager!(self.item_nft)
+                    .mint_non_fungible(&NonFungibleId::random(), weapon_data));
+            
+            self.system_vault.authorize(||
```
0xOmarA, line 601: Can this be minimized so that `self.system_vault.authorize` is not repeated multiple times? Why not something like:

```rust
self.system_vault.authorize(|| {
   gold.burn();
   resource1.burn();
   resource2.burn();
}
```

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+            else {
+                String::from("Pants")
+            };
+            let hp = dec!("5") * self.seed_50();
+            let def = dec!(".05") * self.seed_50();
+            let hp_bonus = dec!("10") * self.seed_50();
+            let def_bonus = dec!(".1") * self.seed_50();
+            let load = dec!(".01") * self.seed_50();
+            let ability_scale = dec!(0) * self.seed_50();
+            if self.seed_10() < dec!(5) {
```
0xOmarA, line 630: Branching off here leads you to repeating the code twice. However, there are ways to simplify this to make sure that code is not being repeated.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+            let atk = dec!(".05") * self.seed_50();
+            let mag = dec!(".05") * self.seed_50();
+            let spd = dec!("5") * self.seed_50();
+            let atk_bonus = dec!(".1") * self.seed_50();
+            let mag_bonus = dec!(".1") * self.seed_50();
+            let spd_bonus = dec!("10") * self.seed_50();
+            let load = dec!(".01") * self.seed_50();
+            let ability_scale = dec!(0) * self.seed_50();
+            let rng = self.seed_3();
+            if rng == 1 {
```
0xOmarA, line 704: Same comment regarding the branching. Can it be made so that the branching either: a) happens at a different point so that there is not much code repetition, b) variables are used so that the branching happens for the variables but does not happen for the core code? 

As you can see, some part of the code are repeated a number of times throughout the if statement branches despite being the same, such as:
```rust
let new_accessory = self.system_vault.authorize(||
    borrow_resource_manager!(self.item_nft)
        .mint_non_fungible(&NonFungibleId::random(), accessory_data));

self.system_vault.authorize(|| 
    gold.burn());
self.system_vault.authorize(||
    resource1.burn());
self.system_vault.authorize(||
    resource2.burn());
self.system_vault.put(key_bucket);
new_accessory
```

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+        pub fn stage(&mut self, 
+            nft_proof: Proof, 
+            gear: Proof, 
+            stage: u64,
+            ) -> (Bucket, Bucket, Bucket) {
+            // Takes system badge for authorization
+            let key_bucket: Bucket = self.system_vault.take(1);
+            // Data from Proofs
+            let mut nft_data: Account = nft_proof.non_fungible().data();
+            let weapon_data: Weapon = gear.non_fungibles()[0].data();
```
0xOmarA, line 795: It feels to me like this method can fail quite often since the order of the items in the `Proof` is hard-coded into the code. This is not how it should be done in Scrypto. 

In my previous comments, I recommended that you split up the three different types of NFTs you have into multiple resources. If you do so, you will no longer need to assume the position of each one of the items in the proof. You would simply have this method take three `Proof`s in a tuple, vector, or as three separate arguments. The three `Proof`s would belong to the three types of items you have.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+            let helmet_data: Armor = gear.non_fungibles()[1].data();
+            let chest_data: Armor = gear.non_fungibles()[2].data();
+            let pants_data: Armor = gear.non_fungibles()[3].data();
+            let gloves_data: Accessory = gear.non_fungibles()[4].data();
+            let belt_data: Accessory = gear.non_fungibles()[5].data();
+            let shoes_data: Accessory = gear.non_fungibles()[6].data();
+            // Getting data of selected stage:
+            let data = self.stage_data.get(&stage).unwrap();
+            let enemy_1_hp: Decimal = data[1].into();
```
0xOmarA, line 804: The `Enemy`, `Reward` and other structs I proposed in a previous comments would be helpful here.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+            assert!(nft_data.stage >= stage.into() || nft_data.stage == stage.into());
+            assert!(gear.resource_address() == self.item_nft,);
+            assert!(helmet_data.part == String::from("Helmet"));
+            assert!(chest_data.part == String::from("Chest"));
+            assert!(pants_data.part == String::from("Pants"));
+            assert!(gloves_data.part == String::from("Gloves"));
+            assert!(belt_data.part == String::from("Belt"));
+            assert!(shoes_data.part == String::from("Shoes"));
+            // Speed = sum of character + gear speed/Speed penality (Tier1 gear gives 1% penalty per item for 6% total penalty)
+            let speed =
```
0xOmarA, line 843: I'm not entirely certain as to what this does. However, can't the math here be simplified to: `(nft_data.speed + gloves_data.speed + belt_data.speed + shoes_data.speed)(1 - helmet_data.weight - chest_data.weight - pants_data.weight - gloves_data.weight - belt_data.weight - shoes_data.weight)`? 

If so, can the math for other operations be simplified as well?

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+            mut defense: Decimal,  
+            mut speed: Decimal, 
+            mut enemy_health: Decimal, 
+            mut enemy_damage: Decimal, 
+            mut enemy_defense: Decimal, 
+            mut enemy_speed: Decimal,
+            ) -> Decimal {
+            let speed_ratio = speed / enemy_speed;
+            loop {
+                let mut damage_given = damage - (enemy_defense / dec!(2));
```
0xOmarA, line 972: What you're trying to do can be done by getting the maximum damage which is lower bound by `1`. Something like:
```rust
let damage_given = std::cmp::max(Decimal::one(), damage - (enemy_defense / dec!(2)));
```
You might want to use this method with the other operations as well.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+                    damage_taken = dec!(1)
+                }
+                let mut priority = speed_ratio * self.seed_50();
+                if health < dec!(0) || health == dec!(0) {
+                    return health
+                }
+                if enemy_health < dec!(0) || enemy_health == dec!(0) {
+                    return health
+                }
+                if priority > dec!(1) {
```
0xOmarA, line 987: I think that `priority` should be handled through a rust `loop` instead of having to code in the logic for each possible `priority`.

games/rpg/src/lib.rs
```diff
@@ -0,0 +1,1483 @@
...
+            );
+            let total_exp = nft_data.exp;
+            let class = nft_data.class;
+            let hp = self.char_hp.get(&class).unwrap();
+            let atk = self.char_atk.get(&class).unwrap();
+            let mag = self.char_mag.get(&class).unwrap();
+            let def = self.char_def.get(&class).unwrap();
+            let spd = self.char_spd.get(&class).unwrap();
+            // Idk if this is good, the level scaling seems wack af lmao
+            match total_exp {
```
0xOmarA, line 1211: It seems to me like this part can be changed to be made more general instead of needing to hardcode all of the information in.

0xOmarA (changes requested):

PR #81, opened 2022-05-17
folded_leverage

PostNutCIarity:
Folded leverage is where a user deposits collateral on a lending platform, borrows against their collateral, re-deposits what they borrowed as additional collateral, borrows against the newly added collateral etc etc until the desired leverage is achieved.

I took the Auto-Lend blueprint example from official Scrypto examples and added a transient token to explore concepts around "folded leverage"

This would allow user to:
1. Deposit collateral to borrow
2. Borrow monies
3. Deposit borrowed monies
4. Borrow additional monies with added borrowed collateral
5. Repeat until desired leverage
6. Repay loans

Currently, it's kinda basic and this is a working prototype. But I want to explore additional use-case with this concept.

### DIFF
--- a/defi/folded_leverage/Cargo.toml
+++ b/defi/folded_leverage/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "auto-lend"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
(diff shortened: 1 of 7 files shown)


0xOmarA:
Hey @PostNutCIarity, is this PR ready for us to review it, or are there any changes coming to it soon? :)

PR #94, opened 2023-02-09
Initial commit

atoumbre:
This is a simple example of a basic single token liquidity pool blueprint. It uses AMM liquidity pool token mechanic to ensure fair distribution of fees based on when liquidity provider deposit or remove tokens.


motosharpley:
Congrats! Your submission has been accepted!

PR #104, opened 2023-03-11
Trust More

kangaderoo:
Read the outline in the readme.


kangaderoo:
This solution is also usable with respect to the mandadory kill-switch requirement made bij the EU.
It should however not be a missuse of the royalty setting, but preferably a build in functionality.

motosharpley:
Congrats your submission has been accepted, Thanks For Contributing!

PR #121, opened 2023-05-18
Access Pass Blueprint

olit123:
A blueprint that facilitates composable access rules.


motosharpley:
Thanks for your submission but closing due to Scrypto Auth system changes in v0.10

PR #123, opened 2023-06-04
Update to Scrypto 0.9.0 and add support for multiple identity providers.

michaelwoodson:
Just wanted to catch this example up to the latest version and play around with Scrypto. New version adds support for multiple identity providers and changes rate to 1000 UBI/month since time is now available.


motosharpley:
thanks for updating!

PR #125, opened 2023-06-20
Scrypto 101 Blueprint Submission: Modified Gumball Machine

smaul0:


motosharpley:
Thank you for your submission. Although you have made some changes to the gumball machine we would like to see an example of a unique blueprint created from scratch to validate that you have gained a complete understanding of the concepts taught in Scrypto 101.

PR #129, opened 2023-08-02
Added uniswap v3 project using scrypto

smaul0:
I wrote Uniswap V3 using Scrypto language. It's static for now as i need BitMath library for making it dynamic but i am facing issue while converting `0x100000000000000000000000000000000` as U256 because of this number is too big to fit in rust default max u128 data type. The exact line number is 120 in `defi/uniswap-v3-scrypto/src` folder `bitmap.rs` file where the issue is present.

### DIFF
--- a/defi/uniswap-v3-scrypto/Cargo.toml
+++ b/defi/uniswap-v3-scrypto/Cargo.toml
@@ -1,0 +1,29 @@
+[package]
+name = "uniswap-v3-scrypto"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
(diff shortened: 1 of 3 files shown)


motosharpley:
Thanks for your contribution however your submission cannot be approved at this time. The current version of scrypto is at v0.11.0 vs this example is still written on v0.9.0, the difference in versions contain breaking changes, there are also still todo items which would need to be fully worked out and implemented correctly in order to accept as a incentivized example. You are welcome to resubmit for evaluation after updating and completing the blueprint functionality if you wish.

PR #132, opened 2023-08-16
Adding extended OneResourcePool example

atoumbre:


motosharpley:
Thank you for the submission. It looks pretty good however to make it more user friendly could you please add to the readme instructions for usage / interacting with the example manifests and a couple valid test cases such as the example .rtm files included. Once those items have been addressed you are welcome to resubmit for the incentive again.

PR #133, opened 2023-12-13
PokerDiceApp

kangaderoo:
Please consider this project for the community examples.

### DIFF
--- a/basic/pokerdicebank/Cargo.toml
+++ b/basic/pokerdicebank/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "pokerdicebank"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+sbor = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.1" }
+scrypto = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.1" }
+
+[dev-dependencies]
+transaction = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.1" }
+radix-engine = { git = "https://github.com/radixdlt/radixdlt-scrypto", tag = "v1.0.1" }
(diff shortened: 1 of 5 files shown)


davidcuellarrdx:
I've had a thorough look at your PR, and overall, it's a commendable effort. The incorporation of multiple component methods, authentication roles, and royalties is a strong foundation.

For improvement, I'd suggest considering the following:
- Test Separation: Following best practices, it would be beneficial to break down the scrypto testing into multiple shorter tests instead of one extensive test. This enhances readability, maintainability, and ease of debugging.

- Blueprint Organization: You might want to explore separating the blueprint into multiple files. A good example to refer to is [the gumball machine](https://github.com/radixdlt/gumball-club/tree/main/scrypto/gumball_club/src), which effectively organizes its structure.

On a positive note, it's evident that you're applying this blueprint in your own fullstack dApp. This not only showcases its practicality but also positions you well for the upcoming Scrypto Challenge next year. The Radix team is eager to see your participation and believes your expertise could contribute significantly.

Keep up the great work!

kangaderoo:
Thanks for the review and the feedback 

As for your suggestions with regards to test in shorter test.
Since I wanted to test the component as it would behave in the application, the state of the component needed to be maintained after each test step. The examples that run smaller test, reinitalize the componennt just before each execution. In this project the vaults and keystore of the component is slowly filled with data required for the next test to succeed.
I also run all the same the test using resim, but since that does not create manifest anymore, using the test env. to generate the manifest is my main reason for this usage.
I assumed that due to the fact the component state needed to be maintained a single big test was the best way not to complicate the test-coding.

 As for your suggestion to split up the code over multiple files. I'll keep this in mind.
 I need to check how that would benefit here, since most methods require either the vault or keystore to be a (locally?) accessable.
 I have no idea how to use/access a keyvaluestore or vault from multiple source files. 

I'll be looking forward to the Scrypt Challenge announcement.

